PR #33, opened 2017-07-27
Fix for Javadoc 

therealluigi:
Addressing https://github.com/alfasoftware/morf/issues/17 plus changes to .gitignore for IntelliJ files

PR #34, opened 2017-09-18
Correct Javadoc on DataSetProducer

alexvinall:
close() is called once at the end of processing.


badgerwithagun:
Thanks @alexvinall.  As I said on #33, we need to sort out those last few build system issues sorted out internally and we can merge these PRs.  Things have been a bit manic!

PR #35, opened 2017-10-04
Travis: disable JDK7 build, pipe Maven stdout to /dev/null to prevent hitting 4MB limit

badgerwithagun:
This fixes the current redness without having to default developers to not having console logging.

Ideally we should pipe to a file and grab that as an artifact: https://docs.travis-ci.com/user/uploading-artifacts/


badgerwithagun:
@wnicholson reviewed over-the-shoulder.

venushka:
Actually piping the output to /dev/null means we don't see what's wrong with the build when it fails. As its working in its current state, I'll have a look and see if we can go anything else to keep important information about the build on the console and still get it to complete without hitting the limit.

Issue #38, opened 2017-10-20
Log basic build output to the console


venushka:
With the change in pull request #35 all of the maven build output being redirected to /dev/null we are unable to see the basic build output like if the tests are running, javadoc generation is producing warnings etc.

badgerwithagun:
Unassigning due to lack of progress

venushka:
Build output is on the console log in Travis now, must be fixed. Not an issue anymore, closing.

Issue #9, opened 2017-06-10
Consider switching logging to slf4j


badgerwithagun:
Currently we force a log4j dependency on applications. There's no need.

badgerwithagun:
@venushka - would value your thoughts on this one.

tsg21:
Yes, I think this is a sensible plan.

> Currently we force a log4j dependency on applications. There's no need.

We don't - commons.logging can be hooked into other logging implementations, but it's nowhere near as widely used as slf4j is.

venushka:
I agree, we should switch to slf4j.

One thing to keep in mind though, we should make sure the test execution within Maven should be provided with a logging implementation (log4j like we already have probably) with the required configuration to ensure the logs from tests are redirected to a file and not be printed to console. Travis builds are limited to 4MB of console output for opensource projects, which is more than adequate for the build, but not including tests logs in console.

badgerwithagun:
Yeah I saw that on your fork - makes sense.

tsg21:
I've had a go at this in a fork. It's a reasonably easy find-replace.

badgerwithagun:
@tsg21 - submit a PR!

tsg21:
Don't see any reason why we'd want to stick to commons-logging?

tsg21:
Anyone like to offer up a reason why we should keep the dependency on log4j?

badgerwithagun:
Nope. Let's just do this. Could you create a PR with commons->slf4j and removal of log4j?

tsg21:
Will do

PR #39, opened 2017-10-20
Reducing console output during build

venushka:
Fix for #38

Also removed sudo flag from travis build configuration as it pins the builds to out-dated version of linux that travis are phasing out.


venushka:
Already useful, there are some JavaDoc warnings in the nuodb module.

`
[WARNING] Javadoc Warnings
[WARNING] Picked up _JAVA_OPTIONS: -Xmx2048m -Xms512m
[WARNING] /home/travis/build/alfasoftware/morf/morf-nuodb/src/main/java/org/alfasoftware/morf/jdbc/nuodb/NuoDBHelper.java:22: warning: no @param for stringOfDouble
[WARNING] public static Double nuoDBConvertStringToDouble(String stringOfDouble) {
[WARNING] ^
[WARNING] /home/travis/build/alfasoftware/morf/morf-nuodb/src/main/java/org/alfasoftware/morf/jdbc/nuodb/NuoDBHelper.java:22: warning: no @return
[WARNING] public static Double nuoDBConvertStringToDouble(String stringOfDouble) {
[WARNING] ^
[WARNING] /home/travis/build/alfasoftware/morf/morf-nuodb/src/main/java/org/alfasoftware/morf/jdbc/nuodb/NuoDBHelper.java:33: warning: no @param for referenceToCast
[WARNING] public static Cast nuoDBCastToString(AliasedField referenceToCast, int length) {
[WARNING] ^
[WARNING] /home/travis/build/alfasoftware/morf/morf-nuodb/src/main/java/org/alfasoftware/morf/jdbc/nuodb/NuoDBHelper.java:33: warning: no @param for length
[WARNING] public static Cast nuoDBCastToString(AliasedField referenceToCast, int length) {
[WARNING] ^
[WARNING] /home/travis/build/alfasoftware/morf/morf-nuodb/src/main/java/org/alfasoftware/morf/jdbc/nuodb/NuoDBHelper.java:33: warning: no @return
[WARNING] public static Cast nuoDBCastToString(AliasedField referenceToCast, int length) {
[WARNING] ^
[WARNING] /home/travis/build/alfasoftware/morf/morf-nuodb/src/main/java/org/alfasoftware/morf/jdbc/nuodb/NuoDBHelper.java:43: warning: no @param for bigDecimal
[WARNING] public static String nuoDBConvertBigDecimalToPlainString(BigDecimal bigDecimal) {
[WARNING] ^
[WARNING] /home/travis/build/alfasoftware/morf/morf-nuodb/src/main/java/org/alfasoftware/morf/jdbc/nuodb/NuoDBHelper.java:43: warning: no @return
[WARNING] public static String nuoDBConvertBigDecimalToPlainString(BigDecimal bigDecimal) {
[WARNING] ^
`

badgerwithagun:
My thinking is that the wish to suppress tests logging is a very Travis-specific issue which shouldn't affect tests run in an IDE, local builds or Alfa's internal CI where having everything in the build log is usually more convenient than scrabbling around downloading build artifacts.

Could we perhaps try and delete all these log config files in the main code but just drop them in during Travis builds?

venushka:
> My thinking is that the wish to suppress tests logging is a very Travis-specific issue which shouldn't affect tests run in an IDE, local builds or Alfa's internal CI where having everything in the build log is usually more convenient than scrabbling around downloading build artifacts.

Its a Maven build specific issue in my opinion as the excessive amount of logs printed in the middle of the build log make is very difficult to spot any warnings in the build such as character encoding warning, javadoc warnings, deprecation warnings, etc. We are already doing this for Alfa internal builds when running with jacoco runner, so this is not any different actually.

> Could we perhaps try and delete all these log config files in the main code but just drop them in during Travis builds?

Good point! Got me thinking of finding a way to use a different log4j configuration when running the Maven build, and did indeed find a way. I've changed all the log4j configuration files in test resources to log to console (previously some of them were writing to files and some to console) and added system property to the surefire plugin to use a different log4j configuration file when running the build in `coverage-per-test` profile, which we only use in CI running with coverage. This would make the logs show in the console when running tests in the IDE and local builds. I think this way the CI will let us know when the test fails and optionally collect junit logs artifacts for investigation and also local builds will remain unchanged.

This pull request now include all the changes.

badgerwithagun:
I'm still not convinced by the idea of shipping Morf with log4j config files included at all - feels like log configuration should be the responsibility of the application (related to #9), not of Morf itself.

This is much better than what we have now though.  Tempted to say let's get this in for now and agonise over the rights and wrongs of our logging setup separately.

venushka:
Actually all the log config fields are in test resources directory, so we are really not shipping them. They are only used for testing. Haven’t checked but I think we only depend on log4j library in test scope as well.

badgerwithagun:
It is in test scope, you're right! OK, great, I'm going to approve this.

Issue #5, opened 2017-06-10
Remove XML handling, Xalan, XPP3 dependencies from morf-core


badgerwithagun:
XML handling is a non-core feature and requires dependencies which should be optional. They should be moved to a separate module, like the `DatabaseType`s.

tsg21:
Hmmm. While I agree it's not core in the strict sense, I do think the XML storage is a pretty essential element. It's like to be pretty essential for much of the testing framework, for example.

Still, it would be nice to remove xpp3.

badgerwithagun:
My thinking is that while the idea of streaming a database into/out of a serialised format *is* core, the choice of doing so to XML is not.

Other choices include compressed bytestreams, JSON or any number of alternatives.  XML is an option. That's fine, but as an option it should be a plugin.

tsg21:
Yes, I understand the thinking, but I wonder if it's decomposition taken too far? We pulled the different DB types out into their own modules because they come with hefty dependencies. I think we can probably re-code the `XmlDataSetProducer` to use `javax.xml.stream.XMLStreamReader` (from core Java) instead of xpp3. If we can do that, is the extraction still worthwhile? I'm not convinced it is - the code is going to be pretty minimal.

badgerwithagun:
The complete list of XML related dependencies includes xalan, xpp3 and xmlpull. If we could redesign to avoid all three dependencies, then that would sort out my primary concern.

My secondary concern is whether any APIs in this area are overly specific to XML as an output format, forcing us into a position where we would have to break compatibility in the future in order to introduce an abstraction, which seems pretty inevitable if someone wants/needs a different format.

That is, I would agree, not a strong enough reason to make the 1.0.0 milestone, but a good enough reason to leave this open in case someone feels like tackling it.

badgerwithagun:
Incidentally, @venushka already pulled out the Excel stuff into a separate submodule with very little difficulty. Instinctively, this should be just as easy. Less effort than trying to refactor it to use a different parser would be.

tsg21:
It's very easy. The API it used to use has essentailly been adopted as the core [StAX standard](https://docs.oracle.com/javase/tutorial/jaxp/stax/why.html)

See branch: remove-xpp3

badgerwithagun:
Sweet. Looks good. Definitely solves the main problem.

I still think this belongs in a side API rather than the core one for maintenance reasons, but I'll have a play with that and see if it can be done without much friction.

[...]

PR #40, opened 2017-10-26
Remove xpp3 dependency

tsg21:
Removes the xpp3 and xmlpull dependencies, in favour of the core-Java Stax API.


badgerwithagun:
Fix for #5

tsg21:
Yeah, they belong in the test tree. Not used to maven's resource structure.

tsg21:
SourceXML is in the right place I think.

They were moved from the integration test project to the core, next to the tests. But I put them in the main tree rather than the test one.

Now fixed.

badgerwithagun:
I've made that change in our fork so I can kick off QA.

badgerwithagun:
@tsg21, this failed Alfa Systems automated tests. It seems to fail reading some XML which was read successfully previously.  Exception:

```
Caused by: java.lang.IllegalStateException: Expecting a tag inside [data] but got [5]
	at org.alfasoftware.morf.xml.XmlPullProcessor.readNextTagInsideParent(XmlPullProcessor.java:117)
	at org.alfasoftware.morf.xml.XmlDataSetProducer$PullProcessorRecordIterator.<init>(XmlDataSetProducer.java:994)
	at org.alfasoftware.morf.xml.XmlDataSetProducer$1$1.<init>(XmlDataSetProducer.java:184)
	at org.alfasoftware.morf.xml.XmlDataSetProducer$1.iterator(XmlDataSetProducer.java:184)
```

Seems to be choking on:

```
<?xml version="1.0" encoding="UTF-8"?><table version="2">
  <metadata name="Foo"/>
  <data>
    <!-- Comment -->
    <record id="40646" />
    <record id="40648" />

    <!-- Comment  -->
    <record id="40640" />
    <record id="40641" />
  </data>
</table>
```

It's possible I've obfuscated this too much and it no longer errors...

tsg21:
OK, fixed that issue, and also removed the entries from the parent pom.xml.

tsg21:
`&#0;` is a character reference to the [unicode null](http://www.fileformat.info/info/unicode/char/0000/index.htm). 

This is invalid xml. [Character references](https://www.w3.org/TR/xml/#wf-Legalchar), [Legal characters](https://www.w3.org/TR/xml/#NT-Char).  Note #0 is disallowed.

`<record foo="A"bar="B" />` is also invalid xml. [Attributes MUST be space-delimited](https://www.w3.org/TR/xml/#NT-STag)

I guess xpp3 was more lenient.

I propose not to make any further changes to Morf here. I think it's those xml documents that need fixing.

badgerwithagun:
Unfortunately the nulls are pervasive in our various snapshots taken for test start positions (591 hits in one module alone).  We could probably fix them all, but I can't do that right now.

In the future I'd push something like this onto the next major release branch, but given where we are, we should just delay this PR until we've had time to get our house in order (and work out how all those nulls made it into automatically generated start positions in the first place).

tsg21:
I think this has uncovered a deeper bug, possibly in `DatabaseDataSetProducer`.

What I think is happening is that we're asking for the string representation of a fixed-width database column. The database is then returning a string with a unicode null character ([which is valid](https://stackoverflow.com/questions/2523284/java-string-replace-and-the-nul-null-ascii-0-character)), which we then merrily write out as xml.

What I am not sure of is:
1. Whether the source column is set to `NULL`, or just has a null character in it.
1. Whether this happens for all DBs, or is just a quirk of one of them.

tsg21:
Plan:

1. Introduce a new xml format version (3) which uses an escape sequence to represent nulls (rather than the invalid char): \0
2. Make Morf read and write the new format by default
3. Write a wrapper for the xml inputstream which (efficiently) transforms v2 format into v3, turning the invalid chars into escaped strings.

tsg21:
Rebased onto master

morf-core/src/main/java/org/alfasoftware/morf/xml/XmlDataSetConsumer.java
```diff
@@ -208,8 +213,28 @@ public void table(final Table table, final Iterable<Record> records) {
...
           for (Column column : table.columns()) {
             String value = getValue(record, column, table.getName());
+
             if (value != null) {
-              rowValueAttributes.addAttribute(XmlDataSetNode.URI, column.getName(), column.getName(), XmlDataSetNode.STRING_TYPE, value);
+              // Escape any null characters - these aren't valid XML so in format version 3 we escape these as \0
+              // We also, consequently, have to escape backslashes as \\
+              // can't use String::replace as it doesn't like null characters
+              int last = 0;
+              StringBuilder escapedValue = new StringBuilder();
```
badgerwithagun: You should set the initial capacity to avoid array resizing: `new StringBuilder(value.length() * 1.5)` perhaps?  i.e. 50% additional space for backslashes?

morf-core/src/main/java/org/alfasoftware/morf/xml/XmlDataSetProducer.java
```diff
@@ -1003,7 +1013,14 @@ public Record next() {
         // Buffer this record
         RecordBuilder result = DataSetUtils.record();
         for (String columnName : columnNames) {
-          result.setString(columnName.toUpperCase(), xmlPullParser.getAttributeValue(XmlDataSetNode.URI, columnName));
+          String attributeValue = xmlStreamReader.getAttributeValue(XmlDataSetNode.URI, columnName);
+
+          // deal with escaping...
```
badgerwithagun: Worth avoiding the double-pass?

morf-core/src/main/java/org/alfasoftware/morf/xml/Version2to3TranformingReader.java
```diff
@@ -0,0 +1,205 @@
...
+        temporary = new char[] {};
+      }
+    }
+
+    // now search for the string we're replacing
+    for (int idx=0; idx<charsRead; idx++) {
+      if (cbuf[off+idx] == nullRefChars[0]) { // look for the ampersand
+        // The first char matches.
+        // Check whether the subsequent chars make up the ref
+        if (isNullCharacterReference(cbuf, off+idx, charsRead-idx)) {
```
badgerwithagun: Add

```
 // NOPMD This is more readable as it is.
```

After this line.

morf-core/src/main/java/org/alfasoftware/morf/xml/Version2to3TranformingReader.java
```diff
@@ -0,0 +1,205 @@
...
+        // copy the entire temporary buffer into the output
+        System.arraycopy(temporary, 0, cbuf, off, temporary.length);
+        charsRead = temporary.length;
+        temporary = new char[] {};
+      }
+    }
+
+    // now search for the string we're replacing
+    for (int idx=0; idx<charsRead; idx++) {
+      if (cbuf[off+idx] == nullRefChars[0]) { // look for the ampersand
```
badgerwithagun: Can we try and extract some of these expressions into descriptive methods - would make it a bit more readable.

morf-core/src/test/java/org/alfasoftware/morf/xml/TestXmlDataSetProducer.java
```diff
@@ -204,6 +245,57 @@ public void testCompositePrimaryKey() {
...
+        fooTable, 
+        table("Foo")
+          .columns(
+            column("id", DataType.BIG_INTEGER).primaryKey())
+          )
+      );
+
+    List<Record> records = ImmutableList.copyOf(producer.records("Foo"));
+    
+    assertEquals("40646", records.get(0).getValue("id"));
```
badgerwithagun: `.getValue()` has been deprecated.  Use `.getLong()`

morf-core/src/test/java/org/alfasoftware/morf/xml/TestXmlDataSetProducer.java
```diff
@@ -15,29 +15,47 @@
...
+import org.alfasoftware.morf.metadata.Table;
 import org.alfasoftware.morf.xml.XmlStreamProvider.XmlInputStreamProvider;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.TemporaryFolder;
+import org.mockito.ArgumentCaptor;
+import org.mockito.Mockito;
+import org.apache.commons.logging.Log;
```
badgerwithagun: Unused imports

badgerwithagun:
This hit a NPE in our integration tests suites. @tsg21 following up.  Missing a test or two ;)

morf-core/src/main/java/org/alfasoftware/morf/xml/Version2to3TranformingReader.java
tsg21: Done.

morf-core/src/main/java/org/alfasoftware/morf/xml/XmlDataSetConsumer.java
tsg21: Done

morf-core/src/main/java/org/alfasoftware/morf/xml/XmlDataSetProducer.java
tsg21: Done

tsg21:
The NPE is now fixed, and more tests have been added. Need to have another go at getting this in.

Issue #27, opened 2017-07-03
TestXmlDataSetConsumer.testIndexOrdering() failing due to line delimiters mismatching


badgerwithagun:
`org.alfasoftware.morf.xml.TestXmlDataSetConsumer.testIndexOrdering()` fails on my personal machine with the two strings mismatching.  They look identical when compared but have a different `.length()`.  I converted the `testIndexOrdering.xml` file to Unix line endings, and the test passes.

I could submit a PR with [this change](https://github.com/badgerwithagun/morf/commits/fixlf) but there must be a reason why this is failing on my machine and not elsewhere.

Thoughts?

venushka:
I think when we serialise XMLs to disk, we only use line endings and not carriage returns `\r`. This test is comparing the serialised version with the XML file in the repo, which is rightly has only line break `\n` characters.

It doesn't fail on mine, so I presume you might have set your [core.autocrlf](https://help.github.com/articles/dealing-with-line-endings/) enabled on your machine. I would really recommend switching it off if this is the case, the VCS shouldn't really change any files based on OS/settings on client, the files checked out must be identical to what's in the repo really.

tsg21:
Looks like we could ensure this works everywhere by using a `.gitattributes` file.

tsg21:
Possible fix: #41

badgerwithagun:
@tsg21 : OK, this has made things even worse.  Now our internal repo and the GitHub master have mismatching linefeeds. I can't change the internal repo to match GitHub or our tests fail.

Is it possible that `testIndexOrdering()` is generating XML based on local-specific linefeeds?  That would mean that when run on Windows it needs to compare with CRLF, but when run on *Nix it needs to compare with LF.

Something not quite right about that explanation, but I'm a bit run ragged at the moment...

badgerwithagun:
@tsg21 , I assume this can now be closed?

tsg21:
Yes. If the tests are passing (they are) and nobody's complaining about local issues (they aren't) we might as well close this one.

PR #41, opened 2017-10-26
Add .gitattributes file to enforce LF line-endings in xml

tsg21:
... so that tests always pass.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,3 @@
+# Set the behaviour for xml files, in case people have core.autocrlf set differently.
+# We need them to have line feeds in order for tests to pass
+*.xml text eol=lf


tsg21:
Trial fix #27 

@badgerwithagun - can you see whether this resolves the issue for you? (it doesn't happen for me locally)

badgerwithagun:
Needs testing on both Win & *Nix

tsg21:
> Needs testing on both Win & *Nix

Not totally convinced it does. It always worked for me and @venushka locally. It just needs to work for you, @badgerwithagun

badgerwithagun:
I don't have this issue anymore since I fixed my local settings. I'm inclined to leave it and just close the issue.  What do you think, guys?

tsg21:
Well, if we don't include this file then some poor engineer with (legitimately) the same settings as you is going to checkout Morf and not be able to run the tests.

I think we should include it. Merge?

badgerwithagun:
Why not. We can always change our minds later. OK, approving

PR #42, opened 2017-10-26
Remove the dependency on H2 database from morf-core

### LINKED ISSUE #18: H2 dependency in morf-core pom looks unnecessary
Shouldn't need H2 until you get up to integrationtest.  @venushka ?

tsg21:
It's clearly wrong.

Added it to morf-h2 (makes sense) and had to add it to morf-testsupport for now, since that's where the H2DatabaseInspector is.

Code changes are almost entirely to remove references to a ResultSet class in H2 that we were using in a test.

Fixes #18

### DIFF
--- a/morf-h2/pom.xml
+++ b/morf-h2/pom.xml
@@ -42,5 +42,9 @@
       <artifactId>log4j</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>com.h2database</groupId>
+      <artifactId>h2</artifactId>
+    </dependency>
   </dependencies>
 </project>
--- a/morf-testsupport/pom.xml
+++ b/morf-testsupport/pom.xml
@@ -44,5 +44,9 @@
       <groupId>junit</groupId>
       <artifactId>junit</artifactId>
     </dependency>
+    <dependency>
+      <groupId>com.h2database</groupId>
+      <artifactId>h2</artifactId>
+    </dependency>
   </dependencies>
 </project>
--- a/morf-core/pom.xml
+++ b/morf-core/pom.xml
@@ -65,10 +65,6 @@
       <artifactId>guice-multibindings</artifactId>
     </dependency>
     <dependency>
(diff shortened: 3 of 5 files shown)


tsg21:
Do we need to remove the h2 reference from the parent pom.xml?

morf-h2/pom.xml
```diff
@@ -42,5 +42,9 @@
       <artifactId>log4j</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
```
badgerwithagun: Generally, we don't introduce compile-time dependencies on JDBC drivers in dialect modules.

Do we think H2 is different?  It is in Maven Central, it's open source - there's no real reason to avoid it.

We need other runtime stuff too, like the inspector, so I'm happy if we all are.

morf-testsupport/pom.xml
```diff
@@ -44,5 +44,9 @@
       <groupId>junit</groupId>
       <artifactId>junit</artifactId>
     </dependency>
+    <dependency>
```
badgerwithagun: This one is more controversial though.  A dirty trick we could pull is to make `DatabaseInspector.inspect()` a `default` method which uses `ServiceLoader` to locate an implementation on the classpath.  That way you can move `H2DatabaseInspector` to `morf-h2` (test tree) but leave `DatabaseInspector` here.

tsg21: That's not a bad idea.

tsg21: Hmmm, but it's in the `testsupport` project, and I am uncomfortable about moving `H2DatabaseInspector` to the h2 `main` tree.

tsg21:
Suggestion: Merge this PR, since it does improve matters (there's no longer a core depn on H2, only from `testsupport`)

We can then have another crack at the `H2DatabaseInspector`.

badgerwithagun:
OK, happy with that

Issue #7, opened 2017-06-10
Make DSLs immutable


badgerwithagun:
Currently, most elements of the SQL DSL are mutable. This is one of those things that occurred organically over time to make the DSL read more cleanly, but which has increasingly looked out of place. It also prevents a number of performance optimisations.

All SQL dialect elements need to be made immutable, allowing us to resolve some unpleasant API side-effects. Once this is done, we can significantly improve caching and performance.

This will have widespread binary compatibility issues, and even text API compatibility in some edge cases.

badgerwithagun:
First stage is available for merge at #43.

Working on the next stage on https://github.com/badgerwithagun/morf/tree/immutable-statements

badgerwithagun:
Second stage now available for merge in #47.

badgerwithagun:
After this, we're going to have to start breaking compatibility.

badgerwithagun:
#47 merged.

badgerwithagun:
@pstefaniak7 is currently working on the last steps in making Alfa's internal code compatible with "immutable mode".  Once that's done we'll flip the default behaviour around and start removing mutable support.

therealluigi:
@pstefaniak7 what is the progress on this one? Finished? Abandoned?

pstefaniak7:
@therealluigi 
No progress on this since last update in 2018.

PR #43, opened 2017-10-30
Equals, hashCode and optional immutability in AliasedField implementations (#7)

badgerwithagun:
First step towards #7.

The idea is to stay 100% compatible, deprecating constructors and other methods that we can't support long term and paving the way to switching to immutability.

Eventually, we will need to break compatibility to _actually_ make everything immutable, since that means that patterns like this will no longer work:

```
AliasedField myField = field("Foo");
myField.as("aliased");
runner.run(select(myField).from("Bar"));
```

However, we have lots of uses of this, so we need to allow a transition.

The  `AliasedField.IMMUTABLE_DSL_ENABLED` system property allows us to simply flick over support for mutability at run-time.   This will allow us to do the work in parallel and then switch over once we have fixed any existing uses.

This is now ready for merge.  More coming in future pull requests.

### DIFF
--- a/morf-core/src/main/java/org/alfasoftware/morf/sql/element/CaseStatement.java
+++ b/morf-core/src/main/java/org/alfasoftware/morf/sql/element/CaseStatement.java
@@ -88,12 +88,6 @@ public class CaseStatement extends AliasedField implements Driver {
   }
 
 
-  @Override
-  protected boolean refactoredForImmutability() {
-    return true;
-  }
-
-
   /**
    * @return the whenConditions
    */
--- a/morf-core/src/main/java/org/alfasoftware/morf/sql/element/CaseStatement.java
+++ b/morf-core/src/main/java/org/alfasoftware/morf/sql/element/CaseStatement.java
@@ -15,13 +15,14 @@
 
 package org.alfasoftware.morf.sql.element;
 
-import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.List;
 
 import org.alfasoftware.morf.util.DeepCopyTransformation;
 import org.alfasoftware.morf.util.ObjectTreeTraverser;
 import org.alfasoftware.morf.util.ObjectTreeTraverser.Driver;
+
+import com.google.common.collect.FluentIterable;
+import com.google.common.collect.ImmutableList;
 
 /**
  * Represents a case select statement.
@@ -31,7 +32,7 @@ import org.alfasoftware.morf.util.ObjectTreeTraverser.Driver;
 public class CaseStatement extends AliasedField implements Driver {
 
   /** When conditions */
-  private List<WhenCondition> whenConditions = new ArrayList<>();
+  private final ImmutableList<WhenCondition> whenConditions;
 
   /** If all the when conditions fail return this default value*/
   private final AliasedField defaultValue;
@@ -45,7 +46,14 @@ public class CaseStatement extends AliasedField implements Driver {
    */
   public CaseStatement(AliasedField defaultValue, WhenCondition... whenConditions) {
     super();
-    this.whenConditions = Arrays.asList(whenConditions);
+    this.whenConditions = ImmutableList.copyOf(whenConditions);
+    this.defaultValue = defaultValue;
+  }
+
+
+  private CaseStatement(String alias, AliasedField defaultValue, ImmutableList<WhenCondition> whenConditions) {
+    super(alias);
+    this.whenConditions = whenConditions;
     this.defaultValue = defaultValue;
   }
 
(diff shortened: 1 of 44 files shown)


badgerwithagun:
This is now ready to be merged.  Future pull requests will do similar sweeps on the higher level entities (e.g. `SelectStatement`.  That's WIP on this branch https://github.com/badgerwithagun/morf/commits/immutable-statements but not ready for a pull request yet.

morf-core/src/main/java/org/alfasoftware/morf/sql/element/FieldLiteral.java
```diff
@@ -42,24 +44,126 @@
...
+    this.value = value;
+    this.dataType = dataType;
   }
 
 
   /**
    * Constructs a new {@linkplain FieldLiteral} with as string source.
    *
    * @param stringValue the literal value to use
+   * @deprecated Use {@link #literal(String)}
```
tsg21: Shouldn't this refer to the Utils class? (There's a lot of these)

morf-core/src/main/java/org/alfasoftware/morf/sql/element/FieldLiteral.java
```diff
@@ -239,8 +366,40 @@ protected AliasedField deepCopyInternal(DeepCopyTransformation transformer) {
...
-    return this;
+    return (FieldLiteral) super.as(aliasName);
+  }
+
+
+  /**
+   * @see java.lang.Object#equals(java.lang.Object)
+   */
+  @Override
+  public boolean equals(Object obj) {
```
tsg21: Elsewhere you're using the Eclipse auto-gen equals (I think), here: EqualsBuilder. Do we want to be consistent?

morf-core/src/main/java/org/alfasoftware/morf/sql/element/NullFieldLiteral.java
```diff
@@ -15,28 +15,34 @@
...
   @Override
-  protected AliasedField deepCopyInternal(DeepCopyTransformation transformer) {
-    return new NullFieldLiteral();
+  protected NullFieldLiteral deepCopyInternal(final DeepCopyTransformation transformer) {
+    return new NullFieldLiteral(this.getAlias());
   }
 
-
-  @Override
-  public String toString() {
```
tsg21: Why has the toString gone?

tsg21: because the ancestor does it for us?

morf-core/src/main/java/org/alfasoftware/morf/sql/element/Cast.java
```diff
@@ -145,4 +161,40 @@ public void drive(ObjectTreeTraverser traverser) {
...
+    if (this == obj)
+      return true;
+    if (!super.equals(obj))
+      return false;
+    if (getClass() != obj.getClass())
+      return false;
+    Cast other = (Cast) obj;
+    if (dataType != other.dataType)
+      return false;
+    if (expression == null) {
```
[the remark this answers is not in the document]
badgerwithagun: They are auto-generated. I don't hold with carefully crafting equals and hashcode methods. Just auto-generate them any time you change the object.

morf-core/src/main/java/org/alfasoftware/morf/sql/element/CaseStatement.java
```diff
@@ -112,8 +125,42 @@ public String toString() {
     for (WhenCondition whenCondition : whenConditions) {
       result.append(whenCondition).append(" ");
     }
+    result.append("ELSE ").append(defaultValue);
```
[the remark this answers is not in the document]
badgerwithagun: It's a fix.  The tests rely on decent toString methods to work.

morf-core/src/main/java/org/alfasoftware/morf/sql/element/FieldLiteral.java
badgerwithagun: Depends whether we think these static methods belong here and whether we actually want to deprecate use of the constructors.  Are they harmful?  Might make life easier to not deprecate them, then there's no value in adding these static methods. Thoughts?

morf-core/src/main/java/org/alfasoftware/morf/sql/element/NullFieldLiteral.java
badgerwithagun: Correct, and in a better way ("NULL" instead of "null" which was a bit misleading).  You've reminded me though that I need to tweak the dialect so a null FieldLiteral is equivalent to a NullFieldLiteral.  This was in this PR and got culled by accident when I was trying to slim it down to a manageable size.

badgerwithagun: Could you elevate this to an review issue?

morf-core/src/main/java/org/alfasoftware/morf/sql/element/FieldLiteral.java
badgerwithagun: Can do.  Bump this up to a review issue if you like and I'll tidy it up.

morf-core/src/main/java/org/alfasoftware/morf/sql/element/NullFieldLiteral.java
badgerwithagun: Fixed

badgerwithagun:
Internal QA passed.  No integration test failures. This is ready for merge.

PR #47, opened 2017-11-19
Immutability/equals/hashCode support for statements

badgerwithagun:
Next step for #7.

Follows on from #43.

This does the same thing for statements as I did previously (in #43) for the lower level elements.  Also introduces a lot more test saturation so we can be sure the assumptions this bakes in continue to be maintained long term.

Once this is in, the next steps are to deprecate and remove all the various cases where mutability is required by existing code so we can start removing some of the mutability support code.


morf-core/src/main/java/org/alfasoftware/morf/sql/AbstractSelectStatement.java
```diff
@@ -125,6 +161,9 @@ protected void addFields(AliasedFieldBuilder... aliasedFields) {
    * @param aliasedFields The fields to add
    */
   protected void addFields(List<? extends AliasedFieldBuilder> aliasedFields) {
+    if (AliasedField.immutableDslEnabled()) {
```
tsg21: Consider pushing this repeated block into a new method on AliasedField: assertImmutableDslEnabled()

morf-core/src/main/java/org/alfasoftware/morf/sql/AbstractSelectStatementBuilder.java
```diff
@@ -0,0 +1,486 @@
...
+
+/**
+ * Base builder for select statements.
+ *
+ * @author Copyright (c) Alfa Financial Software 2017
+ *
+ * @param <U> The statement type.
+ * @param <T> The builder type.
+ */
+public abstract class AbstractSelectStatementBuilder<U extends AbstractSelectStatement<U>, T extends AbstractSelectStatementBuilder<U, T>> implements Builder<U> {
```
tsg21: Why is this abstract?

morf-core/src/main/java/org/alfasoftware/morf/sql/SelectStatement.java
```diff
@@ -42,77 +44,104 @@
...
+ *    .from([table])                       = SELECT [fields] FROM [table]
+ *    .innerJoin([table], [criterion])     = SELECT [fields] FROM [table] INNER JOIN [table] ON [criterion]
+ *    .leftOuterJoin(...)
+ *    .where([criterion])                  = SELECT [fields] FROM [table] WHERE [criterion]
+ *    .orderBy([fields])                   = SELECT [fields] FROM [table] ORDER BY [fields]
+ *    .groupBy([fields])                   = SELECT [fields] FROM [table] GROUP BY [fields]
+ *      .having([criterion])               = SELECT [fields] FROM [table] GROUP BY [fields] HAVING [criterion]
+ *    .union([SelectStatement])            = SELECT [fields] FROM [table] UNION [SelectStatement]
+ *    .unionAll([SelectStatement])         = SELECT [fields] FROM [table] UNION ALL [SelectStatement]
+ *    .build()</pre></blockquote>
```
tsg21: Hmmm, hasn't this reformatting lost some of the flow? It wasn't supposed to imply that each call was chained, but now it looks like they are.

morf-core/src/main/java/org/alfasoftware/morf/sql/element/Criterion.java
```diff
@@ -554,23 +554,11 @@ public String toString() {
...
-      .dispatch(getSelectStatement())
-      .dispatch(getCriteria());
-  }
-
-
-  @SuppressWarnings("rawtypes")
-  private void visitValue(ObjectTreeTraverser traverser,Object value) {
-    if (value instanceof AliasedField) {
-      traverser.dispatch( (AliasedField)value);
-    } else if (value instanceof Iterable) {
```
tsg21: We've lost this Iterable handling. Not required any more?

morf-core/src/main/java/org/alfasoftware/morf/sql/element/Join.java
```diff
@@ -187,6 +187,9 @@ public SelectStatement getSubSelect() {
    */
   @Deprecated
   public Join on(Criterion onCondition) {
+    if (AliasedField.immutableDslEnabled()) {
```
tsg21: This block again...

morf-core/src/main/java/org/alfasoftware/morf/sql/AbstractSelectStatement.java
badgerwithagun: Couldn't find it repeated anywhere, but moved it there anyway.

morf-core/src/main/java/org/alfasoftware/morf/sql/AbstractSelectStatementBuilder.java
badgerwithagun: How could it not be?

morf-core/src/main/java/org/alfasoftware/morf/sql/SelectStatement.java
badgerwithagun: Not sure I follow?

morf-core/src/main/java/org/alfasoftware/morf/sql/element/Criterion.java
badgerwithagun: It's now handled by the traverser (always should have been).  `visitValue(transfers, getValue())` has become `traverser.dispatch(value)`.  That handles and traverses through iterables automatically.

morf-core/src/main/java/org/alfasoftware/morf/sql/element/Join.java
badgerwithagun: Done

badgerwithagun:
Rebased over master ready for merge

badgerwithagun:
Internal QA complete

Issue #14, opened 2017-06-29
Finish refactoring out Guice dependency


badgerwithagun:
We should not have a Guice dependency from morf-core.  Work is in progress to factor this out into a new morf-guicesupport module.

harrisric:
related to #70

PR #48, opened 2017-11-22
Create guicesupport module

badgerwithagun:
Works towards #14.

Moves the `guicesupport` package to a new Maven module, and migrates the JIT bindings for `SqlScriptExecutor` and `TableLoaderBuilder` to there, removing `@Inject` and `@ImplementedBy`.

This is broadly the same pattern we will follow to remove the Guice stuff elsewhere (effectively, change everything to Dagger style bindings).  Will hack away at those as a series of future PRs.


badgerwithagun:
@wnicholson, are you able to review this one?  Otherwise I can assign it elsewhere.

wnicholson:
Looks good to me - merging + two tiny tweaks.

badgerwithagun:
This broke builds at Alfa and had to be reverted.  The main problem appears to be various builds we have which assume there are JIT bindings in place.  This either means that they blow up trying to *replace* said bindings, or simply assume the default bindings are in place.

badgerwithagun:
This PR is in a merged state so I'll submit a second one at some point.

PR #51, opened 2017-11-24
Replace commons-logging and log4j with slf4j

tsg21:
This removes the dependencies on log4j and commons-logging, replacing them everywhere with slf4j.


badgerwithagun:
Hm, hard to tell what's going on on this PR.  Lots of commits fast-forwarded from master...

tsg21:
Yeah, I think I messed that up. Is there a way to squash all the commits into one? I don't really want any of those. It's only the diff from master that we care about.

badgerwithagun:
Just rebase your branch against master, picking the commits you want.

badgerwithagun:
What you wanted was `no-ff`: https://stackoverflow.com/questions/9069061/what-is-the-difference-between-git-merge-and-git-merge-no-ff

jsimlo:
Closing conflicting PR: I have created an internal issue aimed at discussing commons-logging and log4j vs slf4j.

badgerwithagun:
Across all Alfa's code? This only refers to Morf, which would by no means be the only external library that has to be mapped from Slf4j to commons

badgerwithagun:
Surely Slf4j is already in Alfa's stack, is used by countless third party libraries and is already mapped through? What's the harm in using it in one more?

Issue #54, opened 2018-02-07
Remove invalid 'append' Oracle hint


franksloan:
The hint 'append' can be removed from our parameterised _INSERT INTO_ statements, this will be ignored by oracle when executed. The [Oracle guide](https://docs.oracle.com/cd/E11882_01/server.112/e41084/sql_elements006.htm#SQLRF50901) states that:

> The APPEND hint instructs the optimizer to use direct-path INSERT with the subquery syntax of the INSERT statement.

Even though APPEND_VALUES is the correct hint for this type of insert it is not appropriate here and would only have performance benefits when executed within a PL/SQL operation.

badgerwithagun:
Pushed to master

PR #55, opened 2018-02-07
Removing invalid append hint from the oracle dialect

franksloan:
The hint is ignored when executed on oracle.

#54


badgerwithagun:
As an Alfa internal change, this one has been promoted via `alfa-integration-test`.

I'll leave this PR for the moment and cherry-pick the change into master once the tests are green.  Won't be able to merge the PR via GitHub.

badgerwithagun:
Promoted via the normal route.

PR #61, opened 2018-10-31
New test which demonstrates the "start here" part of the Wiki, and fix for it

badgerwithagun:
As of #46 this test fails.


badgerwithagun:
Note that the build failure is going to be fixed by #62.  Please ignore it and merge the PR if the changes look good.

Issue #56, opened 2018-02-12
Publish snapshots to oss.sonatype.org


badgerwithagun:
We're a little way off publishing releases, but not having builds available for developers to pull in using Maven is preventing anyone playing with it (myself included ;)).

Shall we set up on sonatype and deploy there?

Alternatively you can self-host on Github: https://stackoverflow.com/questions/14013644/hosting-a-maven-repository-on-github

badgerwithagun:
@venushka @wnicholson thoughts?

venushka:
Actually I wanted to do this as soon as we got morf on to Github :)

Its actually quite simple to do, the instructions are here: http://central.sonatype.org/pages/ossrh-guide.html

Code wise, we'll need to put back the <distributionManagement> and <scm> blocks back in to the POMs with the appropriate URLs, and we can get Git **master** builds to publish snapshots with every build.

Releases are just a bit more work, I haven't done one of those in Travis, but we can give this a go when we have a repository to publish to. When we publish a release, it needs to be published to a **staging** repository, and promoting it to **release** repository is a manual step from the Nexus UI after we confirm that what we've built is correct. So we can safely try out building releases without actually releasing.

wnicholson:
Agreed. We'll need to apply for a namespace ({{org.alfasoftware}}?) and comply with http://central.sonatype.org/pages/requirements.html.

badgerwithagun:
Anyone fancy pushing that forward?

venushka:
I can look at the making the project compliant with the requirements and getting it to publish.

@wnicholson, would you mind creating the JIRA project and requesting our namespace from Sonatype as mentioned on http://central.sonatype.org/pages/ossrh-guide.html?

wnicholson:
(finally) done - with Sonatype to check & approve: https://issues.sonatype.org/browse/OSSRH-42285

badgerwithagun:
@venushka : this has been approved. What's next?

PR #62, opened 2018-10-31
Publishing to Maven Central

venushka:
Implements #56

I've updated the POM to be compliant with the requirements stated on Maven Central publishing guide and changed the build to publish to Maven Central automatically.

The build has been changed so that,
- builds of **master** now has two jobs,
  -  _test_ job that runs tests and publishes coverage data to SonarCloud
  - _deploy_ job that runs tests and publishes to Maven Central
     - This will publish a SNAPSHOT build if the pom.xml has a -SNAPSHOT version
     - Otherwise, will publish a release*
- builds of **pull requests** or **branches** no has one job,
  - _test_ job that runs the tests. 
- builds of **any fork**, does not matter if its master/pull request/branch, they will all has one job,
  - _test_ job that runs the tests.

I've successfully published two snapshots to https://oss.sonatype.org/content/repositories/snapshots/org/alfasoftware/. I've since change the travis.yaml file to not allow publishing from forks, so going forward, all publishing will only happen from this repository's master.

I have not tested publishing a release as I'm not sure if we want to publish a release now or after checking the open item for the initial release. This should work given the authentication, signing and publishing all works.

\* When publishing a release, we'll need to push a change with the fixed version number in the pom.xml and let a build run to publish. We probably should use the Maven release plugin, but that complicates the publishing process, so we'll probably revisit this again at some point? (using the release plugin will give us benefits like failing releases when using snapshots, but I guess this project is small enough to manage these ourselves?)

Also need to remember that publishing a release itself is not enough, we'll have to **push the release from staging** environment as described here: https://central.sonatype.org/pages/releasing-the-deployment.html

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,7 @@ jobs:
       name: "Deploy"
       script:
         - mvn clean deploy --settings .maven-settings.xml -B -U -Prelease
-    - if: NOT branch = master
+    - if: NOT branch = master OR NOT repo = alfasoftware/morf
       name: "Test"
       script:
         - mvn clean test -B -U
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,17 +7,17 @@ addons:
       secure: "Rk0cQ9JL2/DB8xrKjaU/m9HEvZllokUZ8c7Q+6g3DHwOsMvZi1OoCdUtDXf9Zl0xtdFPMvoWsGPeZ6+sZK9evHscvRH83hVMTrTabWK4+QL660QeMXboBprwuBIpAI0XBCfNcslsT3zPd/e4v35FHPdvRr1yOQlibrjEgs1MZo6/LQ7QgRNJAHsskiqOOJS6LLDHceYJEW7wd/y7R0BvXoxch6xQhDOFvowoc22/cEyUVCSCMMas74hD//veSXdHt0CNOY+H+/PAzAYsbM0rkJnrysqO8UgLZwpnP/Mjsoi0InoisvriSl9MYgQwMAfcrtVxxR94vt4Pu62LXsnt6ufGfwFBrdTh1Rna05fzTAmm9bwywIR2UsMwNeM4iDoz+Rc2ImnwMKRipn5lib5kCOlwqjTv6NVi0oV5XEzTfU51HIrQMmmsq60YwMZgcYwckHoVVOojv29dAjLe0Z/LbfUH8MHprt8/MdWA8ir2egxHHW3dXVY11MgDKzQ2RyYb1uPNJgkVfZDm7uHDcj13dFR61Gy1mTdrtpnWx1ojU2IJQ4oKuGb04RbqQ5lCq7dgsHlaeaAJFMlcn1I6IOk8t6QkYy/3vhSXgkZpQAO3K5Zp4aaGZD90sYdQO3JQNCe2S8mhgEGOqsAdxytFygCaMxOKG5NV9qxVilkzfGPvmqM="
 jobs:
   include:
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -368,24 +370,83 @@
...
+  <licenses>
+    <license>
+      <name>The Apache License, Version 2.0</name>
+      <url>http://www.apache.org/licenses/LICENSE-2.0.txt</url>
+    </license>
+  </licenses>
+  
+  <developers>
+    <developer>
+      <name>Graham Crockford</name>
```
badgerwithagun: We actually have a complete developer list!  Should at least have more than one of us here.

.travis.yml
```diff
@@ -5,15 +5,26 @@ addons:
...
+    - if: branch = master AND repo = alfasoftware/morf AND NOT type = pull_request
+      name: "Test with coverage"
+      script:
+        - mvn clean org.jacoco:jacoco-maven-plugin:prepare-agent test sonar:sonar -Pcoverage-per-test -Dsonar.branch.name=master -B -U
+    - if: branch = master AND repo = alfasoftware/morf AND NOT type = pull_request
+      name: "Deploy"
+      script:
+        - echo $GPG_SECRET_KEYS | base64 --decode | $GPG_EXECUTABLE --import
+        - echo $GPG_OWNERTRUST | base64 --decode | $GPG_EXECUTABLE --import-ownertrust
+        - mvn clean deploy --settings .maven-settings.xml -B -U -Prelease
```
badgerwithagun: Can we not create a profile which does both the testing with coverage _and_ release and avoid the two separate `mvn` calls?

badgerwithagun:
This looks fantastic, @venushka.  *So* stoked that we've finally got this one out of the way. Thanks. Just a few comments to address.

venushka:
> So, to be clear, please correct me if I'm wrong:
> 
> * Every push to master will trigger tests, Sonar, and, under normal circumstances, a SNAPSHOT deploy to Central.
> * If we want a release, we just push a pom.xml change with a non-SNAPSHOT version number. This will cause a release deploy to Central. We should then immediately follow it with a push which restores the SNAPSHOT qualifier and it's back to business as usual
> * Presumably, we will do those pom changes locally using maven-release-plugin.
>
> If that's right, it makes sense to me.

Yes for the first two, for the third one, we should be able to use the release plugin's prepare stage and push the changes than requesting the release plugin to do the deploy part as the credentials for publishing are only on Travis.

.travis.yml
venushka: We could, thought it would be better to keep them separate to reduce the build time (as they run in parallel) and the build output so that we don't exceed the allowed limit on Travis. 

The test + coverage job now runs mvn test (not package), so it doesn't do the unnecessary packaging work it used to do before, where as the deploy task focuses on essentials for publishing. Typically people add -DskipTests to deploy commands, which we could as well, but I felt that's probably not a good idea (but we have the ability to do that if we need to).

I'm not attached to either approaches really :) @badgerwithagun, @wnicholson, thoughts?

pom.xml
venushka: I've added everyone who has contributed on GitHub based on https://github.com/alfasoftware/morf/graphs/contributors in alphabetical order. Hope I got all the details right :)

.travis.yml
badgerwithagun: Meh, this is fine for now. We can always change our minds!

pom.xml
badgerwithagun: I think just the key maintainer list is enough.  Otherwise we'll be changing this list constantly.

badgerwithagun: Changed it myself

PR #73, opened 2018-11-19
Support for direct URLs in H2 and MySQL. 

badgerwithagun:
Needs refinement and testing but demonstrates an approach. Got my use case working for the time being.
 
Thoughts?

### DIFF
--- a/morf-core/src/main/java/org/alfasoftware/morf/jdbc/UrlConnectionResourcesBean.java
+++ b/morf-core/src/main/java/org/alfasoftware/morf/jdbc/UrlConnectionResourcesBean.java
@@ -19,6 +19,7 @@ package org.alfasoftware.morf.jdbc;
 import java.sql.Connection;
 import java.sql.DriverManager;
 import java.sql.SQLException;
+import java.util.regex.Pattern;
 
 import javax.sql.DataSource;
 import javax.sql.XADataSource;
@@ -40,6 +41,8 @@ import com.google.common.base.Preconditions;
 public class UrlConnectionResourcesBean implements ConnectionResources {
 
   private static final Log log = LogFactory.getLog(UrlConnectionResourcesBean.class);
+  
+  private static final Pattern REMOVE_CREDENTIALS = Pattern.compile("//(.*)@");
   
   private final String url;
   private final String databaseType;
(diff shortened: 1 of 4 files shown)


morf-h2/src/main/java/org/alfasoftware/morf/jdbc/h2/H2.java
```diff
@@ -141,4 +149,36 @@ public boolean matchesProduct(String product) {
   public Optional<JdbcUrlElements> extractJdbcUrl(String url) {
     return Optional.empty();
   }
+
+  
+  @Override
+  public Optional<UrlConnectionResourcesBean> urlToConnectionResources(String url) {
```
tsg21: This is a utility method, right? It doesn't @Override something. Should it be static?

morf-mysql/src/main/java/org/alfasoftware/morf/jdbc/mysql/MySql.java
```diff
@@ -144,4 +145,30 @@ public boolean matchesProduct(String product) {
 
     return Optional.of(connectionDetails.build());
   }
+  
+
+  @Override
+  public Optional<UrlConnectionResourcesBean> urlToConnectionResources(String url) {
```
tsg21: Same for this - should it be static?

morf-core/src/main/java/org/alfasoftware/morf/jdbc/UrlConnectionResourcesBean.java
```diff
@@ -0,0 +1,247 @@
...
+public class UrlConnectionResourcesBean implements ConnectionResources {
+
+  private static final Log log = LogFactory.getLog(UrlConnectionResourcesBean.class);
+  
+  private final String url;
+  private final String databaseType;
+  private final String databaseName;
+  
+  private String schemaName;
+  private int transactionIsolationLevel = Connection.TRANSACTION_READ_COMMITTED;
```
tsg21: I would expect this to default to "whatever is supplied by the driver", rather than read-committed specifically (which is a bit of an Alfaism)

morf-h2/src/main/java/org/alfasoftware/morf/jdbc/h2/H2.java
badgerwithagun: It does override the method in `DatabaseType`.

morf-mysql/src/main/java/org/alfasoftware/morf/jdbc/mysql/MySql.java
badgerwithagun: Ditto

morf-core/src/main/java/org/alfasoftware/morf/jdbc/UrlConnectionResourcesBean.java
badgerwithagun: Mmm, could just make it optional.

jsimlo:
@badgerwithagun Is this still relevant to you, do you want to proceed with getting this merged? Some diffs do not seem to diff very nicely..

badgerwithagun:
@jsimlo this was a clean rebase. No conflicts.

PR #75, opened 2018-11-21
Connector driver 8 support

badgerwithagun:
Some initial testing with the Connector/J 8+ driver highlighted a few issues which were easily fixed and shouldn't affect compatibility with the old driver.

There's still a warning about using a deprecated driver classname on startup, but that's less safe to resolve for the moment.

### DIFF
--- a/morf-core/src/main/java/org/alfasoftware/morf/jdbc/DatabaseMetaDataProvider.java
+++ b/morf-core/src/main/java/org/alfasoftware/morf/jdbc/DatabaseMetaDataProvider.java
@@ -223,7 +223,7 @@ public class DatabaseMetaDataProvider implements Schema {
     try {
       final DatabaseMetaData databaseMetaData = connection.getMetaData();
 
-      try (ResultSet primaryKeyResults = databaseMetaData.getPrimaryKeys(null, schemaName, tableName)) {
+      try (ResultSet primaryKeyResults = databaseMetaData.getPrimaryKeys(schemaName, schemaName, tableName)) {
         while (primaryKeyResults.next()) {
           columns.add(new PrimaryKeyColumn(primaryKeyResults.getShort(5), primaryKeyResults.getString(4)));
         }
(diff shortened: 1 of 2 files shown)


jsimlo:
@badgerwithagun  Is this still relevant to you, do you want to proceed with de-conflicting this and getting it merged? DatabaseMetaDataProvider has been reworked in the meantime, changes will need to be re-applied to the new code.

badgerwithagun:
I'll take a look.

badgerwithagun:
@jsimlo I could do a blind rebase but have nothing to test this against anymore since my code is now on a highly divergent fork.

Presumably this is fixed already anyway? Connector/J 8 has been out for 3-4 years...

morf-core/src/main/java/org/alfasoftware/morf/jdbc/DatabaseMetaDataProvider.java
```diff
@@ -223,7 +223,7 @@ public boolean isTemporary() {
     try {
       final DatabaseMetaData databaseMetaData = connection.getMetaData();
 
-      try (ResultSet primaryKeyResults = databaseMetaData.getPrimaryKeys(null, schemaName, tableName)) {
+      try (ResultSet primaryKeyResults = databaseMetaData.getPrimaryKeys(schemaName, schemaName, tableName)) {
```
jsimlo, line 226: I do not think we want to use the _catalog_ parameter for all dialects.

morf-core/src/main/java/org/alfasoftware/morf/jdbc/DatabaseMetaDataProvider.java
```diff
@@ -588,13 +588,12 @@ protected void readTableNames() {
     try {
       final DatabaseMetaData databaseMetaData = connection.getMetaData();
 
-      try (ResultSet tables = databaseMetaData.getTables(null, schemaName, null, TABLE_TYPES)) {
+      try (ResultSet tables = databaseMetaData.getTables(schemaName, schemaName, null, TABLE_TYPES)) {
```
jsimlo, line 591: I do not think we want to use the _catalog_ parameter for all dialects.

PR #79, opened 2019-09-23
Bumping up version of h2 database

marcellor:
We need to upgrade H2 database dependency to version 1.4.199 to add support for row level locking (SELECT FOR UPDATE).


marcellor:
Closing PR.

PR #82, opened 2019-11-04
Fix build on Java 11

badgerwithagun:
Note that this still has to be run with `-Dfindbugs.skip=true` as findbugs is not Java 11 compatible.

Switching to spotbugs is easy.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,4 @@ target
 *.iml
 .idea
 **/ivy-ide-settings.properties
+**.versionsBackup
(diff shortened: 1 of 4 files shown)


badgerwithagun:
Partly fixed by #84 but this adds the necessary jaxb imports

wnicholson:
Hi @badgerwithagun, I think this one can close: JDK 11 builds have been running since #123.

Findbugs -> Spotbugs is an excellent point - I've created #170 to track that.

wnicholson:
Closing this now, happy to re-open if needed.

PR #83, opened 2019-11-04
Bump h2

badgerwithagun:


badgerwithagun:
This change was fully subsumed by #84

PR #91, opened 2020-01-10
Copy the licence into jars when deploying

wnicholson:
It's helpful for the released jars to all contain the licence.

Seeing as we always deploy morf-parent, I've used `${project.baseDir}/..` to get 'up' one level to the root.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -242,7 +242,6 @@
         <artifactId>morf-sqlserver</artifactId>
         <version>${project.version}</version>
       </dependency>
-
       <dependency>
         <groupId>commons-codec</groupId>
         <artifactId>commons-codec</artifactId>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -395,6 +394,15 @@
         <gpg.executable>gpg2</gpg.executable>
       </properties>
       <build>
+        <resources>
```
badgerwithagun: This looks like it's going to attempt to add the licence in _every_ JAR (but presumably only work successfully for the parent) right?

wnicholson: Correct - every submodule project will produce a JAR containing the LICENSE (which should, IMO, be included in every one of them)

Using `${project.baseDir}/..` means it will only work for all the submodules (they will look up one level and find the parent). The parent project packages as POM so is unaffected - if it did try to build a JAR, it would fail to find this resource in the level above the root. There's not, as far as I can find, a convenient variable for "the root project folder" (e.g. see [this SO question](https://stackoverflow.com/questions/3084629/finding-the-root-directory-of-a-multi-module-maven-reactor-project)).

If I run `deploy` locally with this change, the Nexus staging plugin creates `my-morf-location/target/nexus-staging/deferred/org/alfasoftware/morf-core/<version>/` and the JAR in there has the LICENSE in /META-INF (before it fails to do the actual deployment).

wnicholson:
#92 has reverted this - it inadvertantly caused other META-INF content to not be packaged.

PR #92, opened 2020-01-24
Revert "Copy the licence into jars when deploying"

capgen628:
Reverts alfasoftware/morf#91

The change has meant an incomplete set of resources are pulled with the release profile and is failing when deploying


wnicholson:
Thanks @capgen628 - I'll need to look at the other META-INF content #91 failed to copy.

morf-core\src\main\resources\org\alfasoftware\morf\metadata\SQL_RESERVED_WORDS.txt, for example.

PR #100, opened 2020-03-04
Fortifying TestDatabaseMetaDataProvider

jsimlo:


morf-integration-test/src/test/java/org/alfasoftware/morf/jdbc/TestDatabaseMetaDataProvider.java
```diff
@@ -444,7 +450,9 @@ protected U featureValueOf(T actual) {
       case "H2":
         return equalTo(tableName.toUpperCase());
       case "MY_SQL":
-        return equalTo(tableName.toLowerCase());
+        return tableNameInCamelCase
```
badgerwithagun: Gotta admit, this looks pretty odd.

Some explanatory commentary would be useful.

jsimlo: MySQL has a config, lower_case_table_names, which controls the lowercase behaviour as opposed to CamelCase. Hmm.. And yet I still have managed to choose the opposite name.. :)

I'll use `SELECT @@GLOBAL.lower_case_table_names`
and rename the field to be more self-explanatory.

PR #105, opened 2020-05-11
Adding MergingDatabaseDataSetConsumer

avalori:
Add a new Database consumer which merges the records to the target DB rather than truncating and inserting as the current DatabaseDataSetConsumer does.


capgen628:
Can you remove all POM changes from the pull request

PR #112, opened 2020-05-15
making travis ci run verify as opposed to test so jacoco coverage is …

BrannanH:
…aggregated


wnicholson:
Makes sense, currently Travis isn't running Sonar on PR builds from forks, so we need to merge to see this in action.

wnicholson:
Coverage is back!

:champagne: 

Thankyou @BrannanH

PR #116, opened 2020-05-20
Lazy table columns info loading

jsimlo:
Fixing DatabaseMetaDataProvider.RealName equals() and hashCode() methods:
They should act like AName's when comparing to any other ANames.


jsimlo:
Note: @harrisric 's commit got mingled with my commit.

PR #123, opened 2020-08-07
JDK 11 compatibility: Fix some Javadoc, and turn off detectJavaApiLink

wnicholson:
Building with JDK 11 works again for me with these changes (nb continue to target Java 8).

Inspired by
http://blog.joda.org/2014/02/turning-off-doclint-in-jdk-8-javadoc.html
https://stackoverflow.com/a/54684795
https://github.com/joel-costigliola/assertj-core/issues/1403#issuecomment-630601376


wnicholson:
I suspect due to the passage of time, there's no overlap between these changes and #82 - FindBugs runs for me (although warns about illegal reflactive access by Groovy), and I didn't need to change any other dependencies.

PR #129, opened 2020-09-23
Fixing Oracle's behaviour when dropping DEFAULT clauses

jsimlo:


morf-integration-test/src/test/java/org/alfasoftware/morf/integration/TestDatabaseUpgradeIntegration.java
```diff
@@ -937,6 +941,56 @@ private void doAddColumn(Class<? extends UpgradeStep> upgradeStep) throws SQLExc
...
+    catch (NoUpgradePathExistsException e) { /* happy path */ }
+  }
+
+
+  /**
+   * Tests adding non-nullable column to the table,
+   * and leaving a default value behind.
+   */
+  @Test
+  public void testAddColumnWithoutDropDefault2() throws SQLException {
```
gilleain, line 974: What's the difference between these tests, I see the extra `defaultValue("OLD")` - is that "WithForgottenDefaultValue" ?

morf-oracle/src/main/java/org/alfasoftware/morf/jdbc/oracle/OracleMetaDataProvider.java
```diff
@@ -740,6 +744,32 @@ private String determineDefaultValue(String columnName) {
     return "";
   }
 
+
+  private String trimActualDefaultValue(final String actualDefaultValue) {
```
gilleain, line 748: It's fine, but this method does a bit more than just trimming. `getActualDefaultValue` or maybe 'parse'?

jsimlo: It started as trimming, but then it kinda of exploded into some fancy stuff.. 👍 
I'll rename it..

morf-integration-test/src/test/java/org/alfasoftware/morf/integration/TestDatabaseUpgradeIntegration.java
jsimlo: testAddColumnWithoutDropDefault1:
- upgrade stops with NoUpgradePathExistsException
- all as expected

testAddColumnWithoutDropDefault2:
- upgrade proceeds and even succeeds, and only then things fail on schema comparison
- this shows how users can "cheat" the upgrade path finder into upgrading
- this actually leaves behind the lurking default value if I am not mistaken

Perhaps some renaming would be nice?

PR #131, opened 2020-10-09
Dsl fixes and omissions

jsimlo:


morf-core/src/main/java/org/alfasoftware/morf/sql/SqlUtils.java
```diff
@@ -129,6 +128,25 @@ public static SelectStatement selectDistinct(AliasedFieldBuilder... fields) {
...
+   * If no fields are specified then this is equivalent of selecting all
+   * fields (i.e. {@code SELECT DISTINCT * FROM x}).
+   *
+   * <p>Usage is discouraged; this method will be deprecated at some point. Use
+   * {@link SelectStatement#select(AliasedFieldBuilder...)} for preference. For
+   * example:</p>
+   *
+   * <pre>SelectStatement.select(myFields).distinct().from(foo).build();</pre>
+   *
+   * @param fields an array of fields that should be selected
```
capgen628, line 142: Let's drop the "an array of" from here.

morf-core/src/main/java/org/alfasoftware/morf/metadata/SchemaUtils.java
```diff
@@ -553,6 +573,15 @@ public TableBuilder columns(Column... columns) {
     }
 
 
+    /**
+     * @see org.alfasoftware.morf.metadata.SchemaUtils.TableBuilder#columns(java.lang.Iterable)
+     */
+    @Override
+    public TableBuilder columns(Iterable<? extends Column> columns) {
+      return new TableBuilderImpl(getName(), columns, indexes(), isTemporary());
```
capgen628, line 578: Can we make columns(Column... columns) call this method since we construct a list in it.

morf-core/src/main/java/org/alfasoftware/morf/metadata/SchemaUtils.java
```diff
@@ -562,6 +591,15 @@ public TableBuilder indexes(Index... indexes) {
     }
 
 
+    /**
+     * @see org.alfasoftware.morf.metadata.SchemaUtils.TableBuilder#indexes(java.lang.Iterable)
+     */
+    @Override
+    public TableBuilder indexes(Iterable<? extends Index> indexes) {
+      return new TableBuilderImpl(getName(), columns(), indexes, isTemporary());
```
capgen628, line 596: Can we make indexes(Index... indexes) call this method since we construct a list in it.

morf-integration-test/src/test/java/org/alfasoftware/morf/integration/TestSqlStatements.java
```diff
@@ -2488,10 +2488,10 @@ private void preparedStatementRecord(SqlDialect sqlDialect, NamedParameterPrepar
     sqlDialect.prepareStatementParameters(
       preparedStatement,
       ImmutableList.of(
-        parameter("column1").type(INTEGER),
-        parameter("column2").type(DECIMAL),
-        parameter("column3").type(STRING),
-        parameter("parameterValue").type(STRING)
+        parameter("column1").type(DataType.INTEGER),
```
capgen628, line 2491: Why are these static imports being removed? I think the DSL usage reads better without the enum name.

jsimlo: Most of the other code used DataType.INTEGER instead of static INTEGER,
which also matched my personal preference (apparently opposite to yours :)),
so I unified the usage in this direction.

I seem to prefer static imports for methods, but not constants. Funny..

morf-core/src/main/java/org/alfasoftware/morf/metadata/SchemaUtils.java
jsimlo: Makes sense.. But then the constructor's code might become untested. I'll try..

morf-core/src/main/java/org/alfasoftware/morf/metadata/SchemaUtils.java
jsimlo: 👍

morf-core/src/main/java/org/alfasoftware/morf/sql/SqlUtils.java
jsimlo: 👍

morf-core/src/main/java/org/alfasoftware/morf/metadata/SchemaUtils.java
jsimlo: Correction: There is no array-based constructor.

jsimlo:
The 11 Code Smells left are not very relevant here.
If SonarCloud understood the class is an abstract test class, it would not flag them up.
Also, we could turn off the "remove this deprecated code someday" one.

jsimlo:
AbstractSqlDialectTest is marked as "101 lines untested".
Which is not true, all those lines get tested, but in other modules.
It seems part of the problem is where the tests are located.

capgen628:
Code changes look good now. We can ignore the code coverage quality gate failure as it mainly refers to one of the testsupport classes.

PR #136, opened 2020-12-31
Unit test failure - text in Japanese - fix WEB-116913

KrzysztofBalejko:
A minor change to ‘TestDatabaseUpgradeIntegration‘ explicitly setting the LocaleRule to ‘UK’ due to the issues reported under WEB-116913. Discussed with Anuj.


jsimlo:
That's a lot of unnecessary line changes, those imports! :)
My main problem is that they do not follow Alfa's default config.
Hence the next Alfa developer will probably change them all back.

PR #139, opened 2021-01-07
SqlDialect clean-up, refactors, improvements for future improvements

jsimlo:
I have cleaned-up SqlDialect and its descendants and unified them a bit.
- Extracted a few methods into the parent where possible.
- Made tiny inconsequential changes to the various dialects to unify behaviour.
- These are mostly in preparation for future channges and improvements.

Reviewing this commit by commit might make more sense than looking at the whole PR diff,
as this is a bigger collection of various refactors and clean-ups, rather than a concise change.

### DIFF
--- a/morf-core/src/main/java/org/alfasoftware/morf/jdbc/SqlDialect.java
+++ b/morf-core/src/main/java/org/alfasoftware/morf/jdbc/SqlDialect.java
@@ -693,8 +693,7 @@ public abstract class SqlDialect {
 
 
   /**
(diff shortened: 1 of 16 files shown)


jsimlo:
I have reviewed the SonarCloud Code Analysis issues and they are false positives.

capgen628 (approved):

capgen628:
The duplicate lines check is a strange one, I can't see anything obvious that warrants a change so will go ahead an merge.

PR #146, opened 2021-02-22
Oracle unicode parameter fix

jsimlo:
NamedParameterPreparedStatement.setString(SqlParameter, String) should use unicode
java.sql.PreparedStatement.setNString(int, String) for Oracle, instead of non-unicode
java.sql.PreparedStatement.setString(int, String).


gilleain (approved):
Yes, makes sense

PR #148, opened 2021-02-26
Fix the signature of AliasedField.in(List<Object>) to AliasedField.in(Iterable<? extends Object>)

jsimlo:
Note: It previously would not compile against ImmutableList<Integer> correctly.


gilleain:
Yes, fine.

PR #155, opened 2021-03-15
postgres-tempTableSchema-fix - Postgres keeps temporary tables in the…

955dan:
Postgres keeps temporary tables in their own special schema. Fix issue whereby schemas are specified when commenting on temporary tables.


morf-postgresql/src/main/java/org/alfasoftware/morf/jdbc/postgresql/PostgreSQLDialect.java
```diff
@@ -207,14 +207,22 @@ protected String getSqlFrom(Cast cast) {
...
+  private String addTableComment(Table table) {
+    StringBuilder sqlBuilder = new StringBuilder("COMMENT ON TABLE ");
+    if (!table.isTemporary()) {
+      sqlBuilder.append(schemaNamePrefix());
+    }
+    return sqlBuilder.append(table.getName())
+      .append(" IS 'REALNAME:[")
+      .append(table.getName())
+      .append("]'")
+      .toString();
```
jsimlo, line 225: Actually, there is org.alfasoftware.morf.jdbc.postgresql.PostgreSQLDialect.schemaNamePrefix(Table)
which already handles temp tables nicely. You sould simply use that here instead of schemaNamePrefix().

This should end up being a one-liner, as we want to move away from explicit StringBuilder usage
and let Java do the StringBuilder-behind-the-scenes optimisation.

jsimlo (changes requested):

jsimlo (approved):
All good :)

PR #163, opened 2021-04-28
Remove dependency on com.google.inject.extensions:guice-multibindings

wnicholson:
This drops the dependency on guice-multibindings - it's been empty since Guice 4.2 when multibindings support moved into the "core" Guice project - see https://github.com/google/guice/wiki/Guice42#changes-since-guice-42

### DIFF
--- a/morf-core/pom.xml
+++ b/morf-core/pom.xml
@@ -65,10 +65,6 @@
       <artifactId>guice-assistedinject</artifactId>
     </dependency>
     <dependency>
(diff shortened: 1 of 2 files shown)


wnicholson:
Guice 5 [stops publishing the empty multibindings extension](https://github.com/google/guice/wiki/Guice501#multibindings), so removing it here unblocks any future Guice upgrade.

gilleain (approved):

PR #171, opened 2021-08-13
Add support for blob length for different dialects

pgghosh:
Fixed bloblength function to use the correct db function based on the dialect.
Also added support for blob literal value evaluation using the appropriate DB function.

Changes made to MySQL, Oracle and PostGreSQL dialects only as these are the only supported dialects as of now.

Blob literals store binary data in HEX format as most DBs support saving data to DB columns in HEX format.

### DIFF
--- a/morf-core/src/main/java/org/alfasoftware/morf/jdbc/SqlDialect.java
+++ b/morf-core/src/main/java/org/alfasoftware/morf/jdbc/SqlDialect.java
@@ -2222,7 +2222,7 @@ public abstract class SqlDialect {
    * @see org.alfasoftware.morf.sql.element.Function#blobLength(AliasedField)
    */
   protected String getSqlforBlobLength(Function function) {
-    return String.format("OCTET_LENGTH(%s)", getSqlFrom(function.getArguments().get(0)));
+    return String.format("LENGTH(%s)", getSqlFrom(function.getArguments().get(0)));
   }
 
 
--- a/morf-core/src/main/java/org/alfasoftware/morf/jdbc/SqlDialect.java
+++ b/morf-core/src/main/java/org/alfasoftware/morf/jdbc/SqlDialect.java
@@ -1761,47 +1761,47 @@ public abstract class SqlDialect {
         throw new IllegalArgumentException("The COUNT function should have only have one or zero arguments. This function has " + function.getArguments().size());
 
       case COUNT_DISTINCT:
-        checkFunctionHasSingleArgument(function);
+        checkSingleArgument(function);
         return getSqlForCountDistinct(function);
 
       case AVERAGE:
-        checkFunctionHasSingleArgument(function);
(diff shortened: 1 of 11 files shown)


morf-core/src/main/java/org/alfasoftware/morf/sql/element/BlobFieldLiteral.java
```diff
@@ -0,0 +1,22 @@
...
+
+/**
+ * Since we store binary data, we will store the values in hex format, since all DBs support inserting
+ * blob data using hex values.
+ * So {@link FieldLiteral#getValue()} will always return the hex encoded string.
+ */
+public class BlobFieldLiteral extends FieldLiteral {
+
+    public BlobFieldLiteral(byte[] bytes) {
+        super("", new BigInteger(bytes).toString(16).toUpperCase(), DataType.BLOB);
```
capgen628, line 16: There's another super constructor that doesn't take the alias

morf-core/src/main/java/org/alfasoftware/morf/sql/SqlUtils.java
```diff
@@ -343,6 +344,28 @@ public static FieldLiteral literal(Character value) {
   }
 
 
+  /**
+   * Constructs a new {@link BlobFieldLiteral} with a byte array source.
+   *
+   * @param value the literal value to use
+   * @return {@link BlobFieldLiteral}
+   */
+  public static BlobFieldLiteral blobliteral(byte[] value) {
```
capgen628, line 353: Could we just call this method blob?

morf-core/src/main/java/org/alfasoftware/morf/sql/SqlUtils.java
```diff
@@ -343,6 +344,28 @@ public static FieldLiteral literal(Character value) {
...
+  }
+
+
+  /**
+   * Constructs a new {@link BlobFieldLiteral} with a String source.
+   *
+   * @param value the literal value to use
+   * @return {@link BlobFieldLiteral}
+   */
+  public static BlobFieldLiteral blobliteral(String value) {
```
capgen628, line 364: Could we just call this method blob?

morf-core/src/main/java/org/alfasoftware/morf/jdbc/SqlDialect.java
```diff
@@ -1966,6 +1965,12 @@ protected String getSqlFrom(Function function) {
     }
   }
 
+  private void checkFunctionHasSingleArgument(Function function) {
```
capgen628, line 1968: checkSingleArgument?

capgen628 (changes requested):

gilleain (approved):

gilleain:
Yes, looks fine.

PR #175, opened 2021-09-21
DeployedViews SQL definitions

jsimlo:


capgen628 (changes requested):

jsimlo:
Test Coverage below 80% reviewed

capgen628:
There's one code smell that can certainly be cleared

jsimlo:
> There's one code smell that can certainly be cleared 

Seems like a Sonar rule that exists just for the sake of existing. How is that ever going to be an issue? :-D

And why is it not complaining about org.alfasoftware.morf.upgrade.db.v5_1_20.UpgradeSteps as well?

Anyways, I would really like to merge all the UpgradeSteps sub-packaged classes into one, to drop the old version-in-the-package-name style, and have all upgrade steps in a single package, or multiple packages based on functionality, not timeboxes. So let's deal with this code smell then.

capgen628:
Sonar has a delta approach of reporting on the quality of the code you've changed. The idea is that is that if it showed a hundred code smells we would juts ignore them, but if it shows us one we are likely to fix it and not add to the existing pile.

PR #180, opened 2021-11-01
Deployed views definitions - Oracle fix - Long view definitions

jsimlo:


jsimlo:
Test coverage is achieved via morf-integration-test module.

capgen628 (approved):
We should get something like this into the central OracleDialect#makeStringLiteral(String). Let's follow-up with that when we have a bit more time.

PR #181, opened 2021-11-03
Fixed OracleDialect.viewDeploymentStatementsAsLiteral(View) chunking.

jsimlo:
OracleDialect.viewDeploymentStatementsAsLiteral(View) mistakenly tried to split a string which was exactly 1 chunk long (exactly 512 characters long view definition), resulting in a single chunk, resulting in concat complaining. View definitions which do not need to be (or cannot be) split into multiple chunks must take the shortcut to the parent implementation.


gilleain (approved):
Yes, fine

PR #182, opened 2021-11-16
Upgrade Mockito dependency to 3.9.0 and refactor tests to use new API

ch359:
- Upgrade Mockito dependency to 3.9.0
- Refactor tests for compatibility
- Move morf-core module to single-threaded test running for compatibility


gilleain:
Looks fine

gilleain (approved):

PR #184, opened 2021-11-18
Increase H2 default lock timeout

jonknight73:
Increase the hardcoded default lock timeout from 60s to 150s after seeing some large inserts timing out.


gilleain (approved):
Yes, looks fine

PR #193, opened 2022-02-16
Graph based upgrade execution

michal-morzywolek:


gilleain (approved):
Went through this PR with Michal, I'm satisfied that the changes do what is intended, are tested thoroughly, and minimise the risk of functional regression.

PR #195, opened 2022-02-18
Adding support for oracle custom hints

i-stevens:


capgen628:
There's a test failure here.

TestNuoDBDialect>AbstractSqlDialectTest.testHints:3191 expected:<SELECT []* FROM SCHEMA2.Foo> but was:<SELECT [/*+  */ ]* FROM SCHEMA2.Foo>

capgen628:
> There's a test failure here.
> 
> TestNuoDBDialect>AbstractSqlDialectTest.testHints:3191 expected:<SELECT []* FROM SCHEMA2.Foo> but was:<SELECT [/*+ _/ ]_ FROM SCHEMA2.Foo>

Probably something in the abstract dialect that's adding the comments for any hints

PR #199, opened 2022-05-20
Replace nul character

jsimlo:


jsimlo:
@gilleain I think I wrongly included your commit in my PR.
Can it stay, or should I try to get rid of it?

gilleain:
> @gilleain I think I wrongly included your commit in my PR. Can it stay, or should I try to get rid of it?

Yes, I think that's fine.

PR #201, opened 2022-05-27
Fixing TableLoader to delegate directly to the original record.

jsimlo:
This avoid invoking the catch-all getValue() via the default interface method.


gilleain (approved):
Looks fine

PR #202, opened 2022-06-01
Allow MySQL to create views over a UNION select

marcellor:
Handle MySQL creatiion of a view over a UNION select


gilleain (approved):
Yes, makes sense

PR #203, opened 2022-06-06
Setting mvn version to 1.7.3.

marcellor:


gilleain (approved):
Necessary to change back the version (/)

PR #204, opened 2022-06-20
Introducing additional technical indexes to Postgres to support nulla…

jsimlo:
…ble columns in unique indexes the same way Oracle does.


jsimlo:
Strange that TestPostgreSQLUniqueIndexAdditionalDeploymentStatements still leaves PostgreSQLUniqueIndexAdditionalDeploymentStatements with such low test coverage.
Several methods explicitly called from the test remain "completely uncovered".
Something must be off with Sonar.

PR #213, opened 2022-09-27
MORF-55 Add support for EXCEPT set operator and DB-link

michalradziwon:


morf-core/src/main/java/org/alfasoftware/morf/sql/element/TableReference.java
```diff
@@ -52,6 +52,12 @@ public class TableReference implements DeepCopyableWithTransformation<TableRefer
    */
   private final boolean temporary;
 
+  /**
+   * If the table exists in another database then it may be possible to access it with an appropriate database link
+   * (not supported by some database platforms)
+   */
+  private final String dblink;
+
```
jsimlo, line 60: Why are you initialising this with empty strings intead of nulls?

Note to self: No schemaName is denoted as null, but no alias is denoted as empty string.
So we have a precedence either way, I have not looked into why empty strings too much.

morf-core/src/main/java/org/alfasoftware/morf/jdbc/SqlDialect.java
```diff
@@ -745,15 +746,14 @@ protected String schemaNamePrefix(@SuppressWarnings("unused") Table table) {
```
jsimlo, line 746: Was it not possible to also drop all the other schemaNamePrefix() methods?

jsimlo (commented):
Good stuff here.
And thanks for changing that schemaNamePrefix() approach.
Just two questions to wrap this up.

morf-core/src/main/java/org/alfasoftware/morf/jdbc/SqlDialect.java
michalradziwon: These calls can't be replaced with `tableNameWithSchemaName(TableReference tableRef)`:

```
    statement.append("INDEX ")
      .append(schemaNamePrefix(table))
      .append(index.getName())
```

```
  public Collection<String> renameIndexStatements(Table table, String fromIndexName, String toIndexName) {
    return ImmutableList.of("ALTER INDEX " + schemaNamePrefix(table) + fromIndexName + " RENAME TO " + toIndexName);
  }
```

Given the goal here was not to completely refactor the code I just left that as it is.

morf-core/src/main/java/org/alfasoftware/morf/sql/element/TableReference.java
michalradziwon: no specific reason for that :roll_eyes: do we want to change that to `null` ?

michalradziwon:
thank you for reviewing Juraj!

jsimlo (approved):

jsimlo:
@capgen628, this can be merged and released at @michalradziwon's convenience.

PR #220, opened 2022-11-03
Added support for ROW_NUMBER as window function

pgghosh:


morf-testsupport/src/main/java/org/alfasoftware/morf/jdbc/AbstractSqlDialectTestWithNoSupportForWindowFunctions.java
```diff
@@ -0,0 +1,41 @@
...
+    @Override
+    protected void verifyNoExceptionForWithoutOrderByClause() {
+        fail("Window function not supported. UnsupportedOperationException should be thrown");
+    }
+
+    private void verifyWindowFunctionNotSupported(Exception e){
+        assertThat(e, Matchers.instanceOf(UnsupportedOperationException.class));
+        assertThat(e.getMessage(), Matchers.containsString("does not support window functions"));
+    }
+}
```
jsimlo, line 41: Given the fact that you can only extend from one abstract class, this is not the best choice.
For example, thee are DbLinks, which are also not supported on H2.

morf-h2/src/test/java/org/alfasoftware/morf/jdbc/h2/TestH2Dialect.java
```diff
@@ -31,7 +31,7 @@
  *
  * @author Copyright (c) Alfa Financial Software 2010
  */
-public class TestH2Dialect extends AbstractSqlDialectTest {
+public class TestH2Dialect extends AbstractSqlDialectTestWithNoSupportForWindowFunctions {
```
jsimlo, line 34: I thought H2 did support WindowFunctions?

pgghosh: Actually speaking all the ones that are showing as unsupported for window functions now support Window functions. 
H2 from 1.4.198 onwards (currently using 1.4.200)
MYSQL : support is provided in 8.0 which is the current version
SqlServer : All supported versions.

Which means I can remove the unsupported exception in all these dialects. It also means that I would not need to do all the hacks to get the tests working.

morf-testsupport/src/main/java/org/alfasoftware/morf/jdbc/AbstractSqlDialectTestWithNoSupportForWindowFunctions.java
pgghosh: Just saw how it is done for dblinks. I was not aware of the assume function used, that could be used to cleverly ignore the test since it is unsupported. At the moment (See other comment), this may not be required as I think I can remove unsupported exception in all the DBs that are currently showing as unsupported.

morf-integration-test/src/test/java/org/alfasoftware/morf/integration/TestSqlStatements.java
```diff
@@ -3231,8 +3231,9 @@ public void testInsertFromSelectWithHints() {
    */
   @Test
   public void testWindowFunction() {
-    assumeTrue(TEST_ONLY_RUN_WITH_WINDOW_FUNCTION_SUPPORT,connectionResources.sqlDialect().supportsWindowFunctions());
+//    assumeTrue(TEST_ONLY_RUN_WITH_WINDOW_FUNCTION_SUPPORT,connectionResources.sqlDialect().supportsWindowFunctions());
 
+      System.out.println("Running test for DB " + connectionResources.sqlDialect().getDatabaseType());
```
jsimlo, line 3236: Forgotten System.out.println here.

jsimlo (changes requested):

gilleain (approved):

jsimlo (approved):

jsimlo:
The Code Smell is irrelevant.

PR #222, opened 2022-12-13
Fix/morf 58 automatic pk index substitution on oracle

farizbenchaoui-alfa:


morf-oracle/src/main/java/org/alfasoftware/morf/jdbc/oracle/OracleDialect.java
```diff
@@ -235,7 +235,12 @@ private String commentOnTable(String truncatedTableName) {
    */
   private String primaryKeyConstraint(String tableName, List<String> newPrimaryKeyColumns) {
```
michal-morzywolek, line 236: This change assumes that the XXXXXX_PK index **NEVER** exists when this method is invoked which is probably true if it's used only with new tables creation. Something to at least consider before committing. :)

michal-morzywolek: If the above is true we probably should think about other cases where the Oracle "feature" can kick in. Like for example changing PK constraint shape in the upgrade step, like adding or removing a column. Or maybe adding a PK constraint to a table which hasn't had it before.

In such situations the "feature" can also be triggered and it seems that it won't be covered by this fix?

@jsimlo

farizbenchaoui-alfa: Changing the shape of the PK is covered by the integration i.e removing column, changing columns or dropping PK. I will add some integration tests to test adding a PK which matches an existing index.

michal-morzywolek: What I mean is not that they would fail because of your change but that your change doesn't fix the issue you are trying to fix for those situations (at least from what I can see).

In other words if you remove PK constraint and add it again you will likely hit the very issue you are trying to fix.

jsimlo (approved):

michal-morzywolek (approved):

PR #228, opened 2023-02-24
Fixing explicit unique index creation backing PK constraints.

jsimlo:
Adding a check validating that such backing index is unique.


jsimlo:
Not what I wanted 👎

PR #237, opened 2023-03-07
Fix/ctas oracle column definitions

ParkinMike:
Fixed a bug with CTAS, Oracle Dialect will now handle null values when creating a new table as select.

### DIFF
--- a/morf-oracle/src/main/java/org/alfasoftware/morf/jdbc/oracle/OracleDialect.java
+++ b/morf-oracle/src/main/java/org/alfasoftware/morf/jdbc/oracle/OracleDialect.java
@@ -63,6 +63,7 @@ import org.alfasoftware.morf.sql.element.Cast;
 import org.alfasoftware.morf.sql.element.ConcatenatedField;
 import org.alfasoftware.morf.sql.element.FieldReference;
 import org.alfasoftware.morf.sql.element.Function;
+import org.alfasoftware.morf.sql.element.NullFieldLiteral;
 import org.alfasoftware.morf.sql.element.SqlParameter;
 import org.alfasoftware.morf.sql.element.TableReference;
 import org.apache.commons.lang3.StringUtils;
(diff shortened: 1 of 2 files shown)


morf-oracle/src/main/java/org/alfasoftware/morf/jdbc/oracle/OracleDialect.java
```diff
@@ -1123,6 +1126,133 @@ public Collection<String> addTableFromStatements(Table table, SelectStatement se
...
+  private String getSqlField(AliasedField currentField, Table table, List<AliasedField> fields) {
+    Column column = table.columns().get(fields.indexOf(currentField));
+
+    return  "CAST("
+                + column.getName()
+                + " AS "
+                + sqlRepresentationOfColumnType(column, false, false, true)
+                + ") AS "
+                + column.getName();
+  }
```
jsimlo, line 1255: the fist `column.getName()` should be the SQL of the `currentField` instead.

morf-oracle/src/main/java/org/alfasoftware/morf/jdbc/oracle/OracleDialect.java
```diff
@@ -1123,6 +1126,133 @@ public Collection<String> addTableFromStatements(Table table, SelectStatement se
...
+   * @param fields - All the fields being added to the select.
+   * @return String value of the CAST.
+   */
+  private String getSqlForNullField(AliasedField currentField, Table table, List<AliasedField> fields) {
+    Column column = table.columns().get(fields.indexOf(currentField));
+    return  "CAST(null AS "
+                + sqlRepresentationOfColumnType(column, false, false, true)
+                + ") AS "
+                + column.getName();
+  }
```
jsimlo, line 1237: I wonder if you needed this only because of the mistake in the cast below.

jsimlo (changes requested):

morf-oracle/src/main/java/org/alfasoftware/morf/jdbc/oracle/OracleDialect.java
ParkinMike: @jsimlo I have fixed this

jsimlo:
Alternative to https://github.com/alfasoftware/morf/pull/241

jsimlo (changes requested):

PR #241, opened 2023-03-23
Fix/ctas oracle column definitions deep copy

ParkinMike:


jsimlo:
Alternative to https://github.com/alfasoftware/morf/pull/237

PR #242, opened 2023-04-06
Add postgres override for row number function

beckyo1:


morf-postgresql/src/main/java/org/alfasoftware/morf/jdbc/postgresql/PostgreSQLDialect.java
```diff
@@ -93,6 +94,25 @@ protected String schemaNamePrefix(Table table) {
...
+
+  @Override
+  protected String getSqlForWindowFunction(Function function) {
+    FunctionType functionType = function.getType();
+    switch (functionType) {
+      case ROW_NUMBER:
+        return "ROW_NUMBER()";
+
+      default:
+        return getSqlFrom(function);
```
jsimlo, line 111: We should call `super.getSqlForWindowFunction(function)`, which will call `getSqlFrom(function)` for us.

jsimlo (approved):

jsimlo:
Note that we do not want to follow the advice on the 2 Code Smells found.
The switch statement is preferable here than an if for future reasons.
And we would not want to turn a FROM into a constant.

PR #246, opened 2023-04-27
SqlDialect.fetchSizeForBulkSelects update

jdm43:
Updating default fetchSizeForBulkSelects and adding support to configure fetchSizeForBulkSelects and fetchSizeForBulkSelectsAllowingConnectionUseDuringStreaming via ConnectionResources.

- ConnectionResources: Adding fetchSizeForBulkSelects and fetchSizeForBulkSelectsAllowingConnectionUseDuringStreaming as new variables configurable via ConnectionResources and inheriting/implementing classes 

- ResultSetIterator: Adding new constructor accepting ConnectionResources to ResultSetIterator where configured fetch size can be set 

- SqlDialect: Updating default SqlDialect.fetchSizeForBulkSelects to 2000, up from 1

- SqlScriptExecutor: Adding new constructor to SqlScriptExecutor to accept a ConnectionResources and set the configured fetch sizes (if set), in all other instances including original constructor, use default from SqlDialect

- SqlScriptExecutorProvider: Updating default injected constructor to inject ConnectionResources, updating get() method to instantiate SqlScriptExecutor with or without ConnectionResources 

- OracleDialect: removing overridding fetchSizeForBulkSelects in favor of new default or environment variables

### DIFF
--- a/morf-core/src/main/java/org/alfasoftware/morf/jdbc/ConnectionResourcesBean.java
+++ b/morf-core/src/main/java/org/alfasoftware/morf/jdbc/ConnectionResourcesBean.java
@@ -443,6 +443,8 @@ public class ConnectionResourcesBean extends AbstractConnectionResources {
     result = prime * result + port;
     result = prime * result + (schemaName == null ? 0 : schemaName.hashCode());
     result = prime * result + statementPoolingMaxStatements;
+    result = prime * result + fetchSizeForBulkSelects;
+    result = prime * result + fetchSizeForBulkSelectsAllowingConnectionUseDuringStreaming;
     result = prime * result + (userName == null ? 0 : userName.hashCode());
     return result;
   }
@@ -472,8 +474,12 @@ public class ConnectionResourcesBean extends AbstractConnectionResources {
       if (other.schemaName != null) return false;
     } else if (!schemaName.equals(other.schemaName)) return false;
     if (statementPoolingMaxStatements != other.statementPoolingMaxStatements) return false;
-    if (!fetchSizeForBulkSelects.equals(other.fetchSizeForBulkSelects)) return false;
-    if (!fetchSizeForBulkSelectsAllowingConnectionUseDuringStreaming.equals(other.fetchSizeForBulkSelectsAllowingConnectionUseDuringStreaming)) return false;
+    if (fetchSizeForBulkSelects == null) {
+      if (other.fetchSizeForBulkSelects != null) return false;
+    } else if (!fetchSizeForBulkSelects.equals(other.fetchSizeForBulkSelects)) return false;
+    if (fetchSizeForBulkSelectsAllowingConnectionUseDuringStreaming == null) {
+      if (other.fetchSizeForBulkSelectsAllowingConnectionUseDuringStreaming != null) return false;
+    } else if (!fetchSizeForBulkSelectsAllowingConnectionUseDuringStreaming.equals(other.fetchSizeForBulkSelectsAllowingConnectionUseDuringStreaming)) return false;
     if (userName == null) {
       if (other.userName != null) return false;
     } else if (!userName.equals(other.userName)) return false;
--- a/morf-core/src/main/java/org/alfasoftware/morf/jdbc/ConnectionResourcesBean.java
+++ b/morf-core/src/main/java/org/alfasoftware/morf/jdbc/ConnectionResourcesBean.java
@@ -474,12 +474,8 @@ public class ConnectionResourcesBean extends AbstractConnectionResources {
       if (other.schemaName != null) return false;
     } else if (!schemaName.equals(other.schemaName)) return false;
     if (statementPoolingMaxStatements != other.statementPoolingMaxStatements) return false;
-    if (fetchSizeForBulkSelects == null) {
-      if (other.fetchSizeForBulkSelects != null) return false;
-    } else if (!fetchSizeForBulkSelects.equals(other.fetchSizeForBulkSelects)) return false;
-    if (fetchSizeForBulkSelectsAllowingConnectionUseDuringStreaming == null) {
-      if (other.fetchSizeForBulkSelectsAllowingConnectionUseDuringStreaming != null) return false;
-    } else if (!fetchSizeForBulkSelectsAllowingConnectionUseDuringStreaming.equals(other.fetchSizeForBulkSelectsAllowingConnectionUseDuringStreaming)) return false;
+    if (!Objects.equals(fetchSizeForBulkSelects, other.fetchSizeForBulkSelects)) return false;
+    if (!Objects.equals(fetchSizeForBulkSelects, other.fetchSizeForBulkSelects)) return false;
     if (userName == null) {
       if (other.userName != null) return false;
     } else if (!userName.equals(other.userName)) return false;
--- a/morf-core/src/main/java/org/alfasoftware/morf/jdbc/ConnectionResourcesBean.java
+++ b/morf-core/src/main/java/org/alfasoftware/morf/jdbc/ConnectionResourcesBean.java
@@ -73,6 +73,8 @@ import com.google.common.io.Resources;
  *  <dt>userName</dt><dd>The JDBC username.</dd>
  *  <dt>password</dt><dd>The JDBC password.</dd>
  *  <dt>statementPoolingMaxStatements</dt><dd>The maximum number of statements to cache in a statement pool.</dd>
+ *  <dt>fetchSizeForBulkSelects</dt><dd>The JDBC Fetch Size to use when performing bulk select operations, intended to replace the default in {@link SqlDialect#fetchSizeForBulkSelects()}.</dd>
+ *  <dt>fetchSizeForBulkSelectsAllowingConnectionUseDuringStreaming</dt><dd>The JDBC Fetch Size to use when performing bulk select operations while allowing connection use, intended to replace the default in {@link SqlDialect#fetchSizeForBulkSelectsAllowingConnectionUseDuringStreaming()}.</dd>
  * </dl>
  *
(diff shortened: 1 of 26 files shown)


morf-core/src/main/java/org/alfasoftware/morf/jdbc/SqlScriptExecutorProvider.java
```diff
@@ -31,19 +32,23 @@ public class SqlScriptExecutorProvider implements Provider<SqlScriptExecutor> {
 
   private final DataSource dataSource;
   private final Provider<SqlDialect> sqlDialect;
+  private ConnectionResources connectionResources;
```
martyntamerlane, line 35: This can be final

morf-core/src/main/java/org/alfasoftware/morf/jdbc/ResultSetIterator.java
```diff
@@ -95,6 +95,38 @@ public ResultSetIterator(Table table, String query, Connection connection, SqlDi
...
+   * Creates a {@link ResultSetIterator} using the supplied SQL query,
+   * applying no further result sorting beyond that contained in the SQL query.
+   *
+   * @param table Meta data for the query's result.
+   * @param query The query to run.
+   * @param connection The database connection to use.
+   * @param connectionResources The connection resources to use.
+   * @param sqlDialect The vendor-specific dialect for the database connection.
+   */
+  public ResultSetIterator(Table table, String query, Connection connection, ConnectionResources connectionResources, SqlDialect sqlDialect) {
```
martyntamerlane, line 108: All of this doesn't need to be duplicated from the other constructor, can we commonalise the code to a single place? I imagined this would be an overloaded constructor that just calls the existing one.

morf-core/src/main/java/org/alfasoftware/morf/jdbc/ConnectionResourcesBean.java
```diff
@@ -87,6 +89,8 @@ public class ConnectionResourcesBean extends AbstractConnectionResources {
   private String schemaName;
   private String userName;
   private int statementPoolingMaxStatements;
+  private int fetchSizeForBulkSelects;
```
martyntamerlane, line 92: Should this be Integer instead of int? Some dialects might have a valid fetch size of 0

martyntamerlane (changes requested):
Looks good, left some suggestions

morf-core/src/main/java/org/alfasoftware/morf/jdbc/SqlScriptExecutorProvider.java
```diff
@@ -31,19 +32,23 @@ public class SqlScriptExecutorProvider implements Provider<SqlScriptExecutor> {
...
+   *                   {@link SqlScriptExecutorProvider} for
    */
   @Inject
-  public SqlScriptExecutorProvider(final DataSource dataSource, Provider<SqlDialect> sqlDialect) {
+  public SqlScriptExecutorProvider(final DataSource dataSource, Provider<SqlDialect> sqlDialect, ConnectionResources connectionResources) {
     super();
     this.dataSource = dataSource;
     this.sqlDialect = sqlDialect;
+    this.connectionResources = connectionResources;
   }
```
jsimlo, line 52: Important: This is a backwards-incompatible change.
Because this is a public constructor, it is possible some code out there is calling this directly.
Don't let the `@Inject` annotation fool you, it can still be used as a regular constructor.

morf-core/src/test/java/org/alfasoftware/morf/jdbc/TestResultSetIterator.java
```diff
@@ -208,6 +214,94 @@ public void testBuildWithEmptyColumnOrdering() throws Exception {
...
+    // When
+    @SuppressWarnings("resource") /* Resources are closed in ResultSetIterator.close()
+                                     automatically when caller attempts to advance past
+                                     the last row in the result set. */
+        ResultSetIterator resultSetIterator = new ResultSetIterator(table, query, connection, Optional.of(connectionResources), sqlDialect);
+
+    // Then
+    assertTrue(resultSetIterator.hasNext());
+
+    resultSetIterator.next();
```
martyntamerlane, line 241: Why do we need to test iteration until an exception is thrown? It's not necessary for testing whether the fetch size was set, so I think these unit tests should be a lot simpler.

morf-core/src/test/java/org/alfasoftware/morf/jdbc/TestSqlScriptExecutor.java
```diff
@@ -52,6 +52,65 @@ public void setUp() throws SQLException {
     when(databaseType.reclassifyException(any(Exception.class))).thenAnswer(invoc -> (Exception) invoc.getArguments()[0]);
   }
 
+  /**
+   * Verify that each {@link SqlScriptExecutor} sets the correct fetch sizes {@link java.sql.PreparedStatement}.
+   */
+  @Test
+  public void testDefaultFetchSizesForSqlScriptExecutor() throws SQLException {
+    verify(sqlDialect, times(1)).fetchSizeForBulkSelects();
```
martyntamerlane, line 60: Shouldn't we instead verify the fetch size being set on the prepared statement?

morf-core/src/main/java/org/alfasoftware/morf/jdbc/ConnectionResourcesBean.java
```diff
@@ -430,6 +472,8 @@ public final boolean equals(Object obj) {
       if (other.schemaName != null) return false;
     } else if (!schemaName.equals(other.schemaName)) return false;
     if (statementPoolingMaxStatements != other.statementPoolingMaxStatements) return false;
+    if (fetchSizeForBulkSelects != other.fetchSizeForBulkSelects) return false;
+    if (fetchSizeForBulkSelectsAllowingConnectionUseDuringStreaming != other.fetchSizeForBulkSelectsAllowingConnectionUseDuringStreaming) return false;
```
jsimlo, line 476: These are boxed Integer values, I am not sure `fetchSizeForBulkSelects != other.fetchSizeForBulkSelects` is correct comparison here. See Objects.equals()

morf-core/src/main/java/org/alfasoftware/morf/jdbc/ResultSetIterator.java
```diff
@@ -85,7 +104,9 @@ public ResultSetIterator(Table table, String query, Connection connection, SqlDi
     try {
       this.statement = connection.createStatement(ResultSet.TYPE_FORWARD_ONLY, ResultSet.CONCUR_READ_ONLY);
       this.statement.setFetchDirection(ResultSet.FETCH_FORWARD);
-      this.statement.setFetchSize(sqlDialect.fetchSizeForBulkSelects());
+      this.statement.setFetchSize((connectionResources.isPresent() && connectionResources.get().getFetchSizeForBulkSelects() != null) ?
+          connectionResources.get().getFetchSizeForBulkSelects() : sqlDialect.fetchSizeForBulkSelects());
```
jsimlo, line 108: `connectionResources.map(ConnectionResources::getFetchSizeForBulkSelects)
    .orElse(sqlDialect.fetchSizeForBulkSelects())`

That will also handle the null-returning `getFetchSizeForBulkSelects()` for you automatically :)

morf-core/src/main/java/org/alfasoftware/morf/jdbc/SqlScriptExecutor.java
```diff
@@ -66,13 +69,37 @@ public class SqlScriptExecutor {
...
+    this.fetchSizeForBulkSelects = connectionResources.getFetchSizeForBulkSelects() != null
+        ? connectionResources.getFetchSizeForBulkSelects() : sqlDialect.fetchSizeForBulkSelects();
+    this.fetchSizeForBulkSelectsAllowingConnectionUseDuringStreaming = connectionResources.getFetchSizeForBulkSelectsAllowingConnectionUseDuringStreaming() != null
+        ? connectionResources.getFetchSizeForBulkSelectsAllowingConnectionUseDuringStreaming() : sqlDialect.fetchSizeForBulkSelectsAllowingConnectionUseDuringStreaming();
+    if (visitor == null) {
+      this.visitor = new NullVisitor();
+    } else {
+      this.visitor = visitor;
+    }
+  }
```
jsimlo, line 102: We are duplicating the null visitor code here. While rather minor thing right now, it feels like a step in the wrong direction.

We could either extract a common private contructor to be called from both of these.
Or we could extract the `if (visitor == null)` logic into a private static method to be called from both of these.
The latter is probably better for the future?

jsimlo (changes requested):

jsimlo:
Note: I do not mind the [2 Code Smells](https://sonarcloud.io/project/issues?id=org.alfasoftware%3Amorf-parent&pullRequest=246&resolved=false&types=CODE_SMELL). Redefining these makes sense to me given the context.

morf-core/src/main/java/org/alfasoftware/morf/jdbc/SqlScriptExecutorProvider.java
jdm43: Discussed with Martyn - leaving this as is, because we’ve left pre-existing public constructors that don’t accept ConnectionResources. Setting this to final will therefore cause compiliation errors.

morf-core/src/test/java/org/alfasoftware/morf/jdbc/TestSqlScriptExecutor.java
jdm43: Discussed with Martyn - leaving this as is as there's not a clean way to assert on this class's private field (fetchSizeForBulkSelects) or the value set on the inner NamedParameterStatement without using something like Powermock.

PR #254, opened 2023-05-18
Adding upgrade script additions provider.

pstefaniak7:


jsimlo (approved):

morf-core/src/main/java/org/alfasoftware/morf/upgrade/UpgradeScriptAdditionsProvider.java
```diff
@@ -0,0 +1,57 @@
...
+        return Collections.emptySet();
+    }
+
+
+    /**
+     * Allows for filtering of script additions.
+     * @param scriptAdditionsPredicate
+     */
+    default void setAllowedPredicate(Predicate<UpgradeScriptAddition> scriptAdditionsPredicate) {
+    }
```
jsimlo, line 31: Just wondering: Perhaps injecting UpgradeScriptAdditionsProviderFactory, which would get the correct Predicate would be better than having a setter, which you need to call from all relevant places, risking some places will be forgotten.

jsimlo (changes requested):

jsimlo (approved):

PR #255, opened 2023-05-24
Add CLOB datatype support to Excel TableOutputter

haddengalfa:


morf-excel/src/main/java/org/alfasoftware/morf/excel/TableOutputter.java
```diff
@@ -79,10 +81,15 @@ class TableOutputter {
    */
   private static final int MAX_EXCEL_COLUMNS = 256;
 
+  /**
+   * The maximum number of CLOB characters supported in an XLS cell.
+   */
+  private static final int MAX_CLOB_CHARACTERS = 1000;
```
jsimlo, line 87: Presumably you get exceptions when you do not trim the string yourself?
Otherwise, I would let XLS to do the trimming to whatever lenght it must trim, rather than here.
If, in the future, XLS increases the limit, we will automatically benefit, if we do not trim ourselves.

jsimlo (changes requested):

jsimlo (approved):

morf-excel/src/main/java/org/alfasoftware/morf/excel/TableOutputter.java
jsimlo: Unless someone explicitly asked for trimmed values, I think it is best to keep them whole.

PR #261, opened 2023-06-14
Terminology updated as suggested

XsPhantomXs:


morf-core/src/main/java/org/alfasoftware/morf/upgrade/UpgradePathFinder.java
```diff
@@ -60,6 +60,14 @@ public class UpgradePathFinder {
...
   private final List<CandidateStep> stepsToApply;
 
+  private final String sourceSchema = "expected schema based on application binaries";
+
+  private final String upgradedSchema = "database schema with upgrades applied";
+
+  private final String reversedSchema = "upgraded database schema with upgrades reversed";
+
+  private final String currentSchema = "current database schema";
+
```
jsimlo, line 70: `private static final String`

jsimlo (changes requested):

morf-core/src/main/java/org/alfasoftware/morf/upgrade/UpgradePathFinder.java
jsimlo: Rename `sourceSchema` to `applicationSchema`. Alternatively `targetSchema` could work?

jsimlo (changes requested):

morf-core/src/main/java/org/alfasoftware/morf/upgrade/UpgradePathFinder.java
XsPhantomXs: Changed to `applicationSchema`

jsimlo (approved):

jsimlo (approved):

PR #262, opened 2023-06-14
Fix/viewchanges correctcase method quadratic time

jmgrc:


morf-core/src/main/java/org/alfasoftware/morf/upgrade/ViewChanges.java
```diff
@@ -73,6 +74,11 @@ public class ViewChanges {
    */
   private final Collection<View> allViews;
 
+  /**
+   * A map representation of allViews in with the view name lowercased as the key and the original text as the value
+   */
+  private final Map<String, String> allViewsMap;
```
jsimlo, line 80: Does this need to be a field?
I think this can be a regular variable within `correctCase()` method, as it is only used once for each `ViewChanges` instance.

jsimlo (changes requested):

jsimlo (approved):

PR #265, opened 2023-06-19
change upgrade so view definitions update based on table existing in …

ParkinMike:
…target schema rather that in DB

### DIFF
--- a/morf-core/src/main/java/org/alfasoftware/morf/jdbc/SqlDialect.java
+++ b/morf-core/src/main/java/org/alfasoftware/morf/jdbc/SqlDialect.java
@@ -3807,7 +3807,7 @@ public abstract class SqlDialect {
 
   /**
    * For some dialects, this casting is required, as the type may not be inferred for every field in the select statement.
-   * @param table the tale to add the casts.
+   * @param table the table to add the casts.
    * @param selectStatement select statements which the casts need adding to.
    * @return SelectStatement with casts.
    */
--- a/morf-core/src/main/java/org/alfasoftware/morf/jdbc/SqlDialect.java
+++ b/morf-core/src/main/java/org/alfasoftware/morf/jdbc/SqlDialect.java
@@ -3807,7 +3807,7 @@ public abstract class SqlDialect {
 
   /**
    * For some dialects, this casting is required, as the type may not be inferred for every field in the select statement.
-   * @param table the talbe to add the casts.
+   * @param table the tale to add the casts.
    * @param selectStatement select statements which the casts need adding to.
    * @return SelectStatement with casts.
    */
--- a/morf-core/src/main/java/org/alfasoftware/morf/jdbc/SqlDialect.java
+++ b/morf-core/src/main/java/org/alfasoftware/morf/jdbc/SqlDialect.java
@@ -3807,6 +3807,9 @@ public abstract class SqlDialect {
 
   /**
    * For some dialects, this casting is required, as the type may not be inferred for every field in the select statement.
+   * @param table the talbe to add the casts.
+   * @param selectStatement select statements which the casts need adding to.
+   * @return SelectStatement with casts.
    */
   protected SelectStatement addCastsToSelect(Table table, SelectStatement selectStatement) {
     SelectStatement statementWithCasts = selectStatement.deepCopy();
--- a/morf-core/src/main/java/org/alfasoftware/morf/upgrade/LoggingSqlScriptVisitor.java
+++ b/morf-core/src/main/java/org/alfasoftware/morf/upgrade/LoggingSqlScriptVisitor.java
@@ -37,7 +37,7 @@ public class LoggingSqlScriptVisitor implements SqlScriptVisitor {
 
   /**
    * Facilitates logging with additional info about the currently upgraded schema position.
-   * @param schemaPosition
+   * @param schemaPosition String containing scheam position.
    */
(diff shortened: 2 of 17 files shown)


morf-core/src/main/java/org/alfasoftware/morf/upgrade/UpgradeSchemas.java
jsimlo, line 1: Since we are promoting this class as an API, perhaps it would be a good idea to have a proper look at this.
Currently it is a classic stateful data holder, getters/setters, etc. Do we actually use the setters for anything?

Could we perhaps drop the setters and make the two fields `private final`, make the whole thing immutable?

And could we make the entire class `public final` as well? This will prevent mocking in tests, but that should be okay.

morf-core/src/main/java/org/alfasoftware/morf/upgrade/LoggingSqlScriptVisitor.java
```diff
@@ -37,7 +37,7 @@ public LoggingSqlScriptVisitor() {
 
   /**
    * Facilitates logging with additional info about the currently upgraded schema position.
-   * @param schemaPosition
+   * @param schemaPosition String containing scheam position.
```
jsimlo, line 40: Typo here.
Actually, let's say `schema position name`.

morf-core/src/main/java/org/alfasoftware/morf/upgrade/ViewChangesDeploymentHelper.java
jsimlo, line 1: ViewChangesDeploymentHelper is effectively a piece of public API, so we cannot just add `UpgradeSchemas upgradeSchemas` to the methods. We need to keep the old methods for backwards compatibility. deprecate them, create new methods as we want them, and finally make the old methods call the new methods.

_(yeah, I know! :))_

morf-core/src/main/java/org/alfasoftware/morf/upgrade/ViewDeploymentValidator.java
```diff
@@ -64,6 +64,8 @@ interface Factory  {
 
     /**
      * Creates a {@link ViewDeploymentValidator} implementation for the given connection details.
+     * @param connectionResources connection resources for the datasource.
+     * @return ViewDeploymentValidator.
      */
     ViewDeploymentValidator createViewDeploymentValidator(ConnectionResources connectionResources);
```
jsimlo, line 70: Do we not need `UpgradeSchemas upgradeSchemas` here?

Ideally we would want to keep the three interfaces relatable to each other, they usually do the job together, whcih makes me wonder why this is happy without checking the schemas.

jsimlo (changes requested):

morf-core/src/main/java/org/alfasoftware/morf/upgrade/UpgradeSchemas.java
```diff
@@ -5,36 +5,33 @@
...
         this.sourceSchema = sourceSchema;
         this.targetSchema = targetSchema;
     }
 
-    public Schema getSourceSchema() {
-        return sourceSchema;
+    public UpgradeSchemas() {
+        this.sourceSchema = null;
+        this.targetSchema = null;
     }
```
jsimlo, line 28: Do we want/need this constructor?

morf-core/src/main/java/org/alfasoftware/morf/upgrade/ViewDeploymentValidator.java
```diff
@@ -31,23 +51,78 @@ public interface ViewDeploymentValidator {
    * first need to be de-registered, to clear any potential remnants, and only then the new view will be created.
    *
    * @param view View being examined. This is a non-existing schema view.
+   * @param upgradeSchemas Source and target schemas used for upgrade.
    * @return This method should return true, if the view can simply be created.
    *         This method can return false to trigger full view re-definition.
    */
-  public boolean validateMissingView(View view);
+  default boolean validateMissingView(View view, UpgradeSchemas upgradeSchemas){
+    return validateExistingView(view);
```
jsimlo, line 59: This should call `validateMissingView`

jsimlo (changes requested):

morf-core/src/main/java/org/alfasoftware/morf/upgrade/ViewDeploymentValidator.java
ParkinMike: updated

jsimlo:
Reviewed the [22 Code Smells](https://sonarcloud.io/project/issues?id=org.alfasoftware%3Amorf-parent&pullRequest=265&resolved=false&types=CODE_SMELL). All expected.

morf-core/src/main/java/org/alfasoftware/morf/jdbc/SqlDialect.java
```diff
@@ -3807,6 +3807,9 @@ private static String getFieldName(AliasedField field) {
 
   /**
    * For some dialects, this casting is required, as the type may not be inferred for every field in the select statement.
+   * @param table the talbe to add the casts.
```
jonknight73, line 3810: talbe is a typo

ParkinMike: Good spot, Corrected.

jsimlo (approved):

Issue #170, opened 2021-07-21
Swap to SpotBugs


wnicholson:
Morf uses [findbugs-maven-plugin](https://gleclaire.github.io/findbugs-maven-plugin/) which uses [FindBugs](http://findbugs.sourceforge.net/) 3.0.1.

FindBugs is a discontinued project, and [SpotBugs](https://spotbugs.github.io/) is its successor. There's a [Maven plugin](https://spotbugs.readthedocs.io/en/latest/maven.html) and [migration is straightforward](https://spotbugs.readthedocs.io/en/latest/migration.html#guide-for-migration-from-findbugs-3-0-to-spotbugs-3-1).

I think Morf should swap to using SpotBugs.

wnicholson:
Having said this is straightforward, I've tried it out and discovered Morf was binding FindBugs to ``verify`` as oppsed to the default, `validate``.

Unfortunately, as FindBugs and SpotBugs both need access to the classes, in the current setup they run and find nothing to do as ``compile`` hasn't happened yet. Moving the plugin config to ``verify`` throws up a selection of errors for ``morf-core`` (the only module currently configured to run FindBugs).

wnicholson:
See https://github.com/wnicholson/morf/tree/switch_to_spotbugs

wnicholson:
I've ported that branch as-is over to alfasoftware so it runs in CI - see https://github.com/alfasoftware/morf/tree/switch_to_spotbugs

25 fails from the [initial run in Travis](https://app.travis-ci.com/github/alfasoftware/morf/builds/244320641):
| Rule                                                | Errors |
|-----------------------------------------------------|-------:|
| NP_PARAMETER_MUST_BE_NONNULL_BUT_MARKED_AS_NULLABLE | 12     |
| NP_PARAMETER_MUST_BE_NONNULL_BUT_MARKED_AS_NULLABLE | 9      |
| NP_NULL_ON_SOME_PATH_FROM_RETURN_VALUE              | 2      |
| NP_BOOLEAN_RETURN_NULL                              | 1      |
| NP_TOSTRING_COULD_RETURN_NULL                       | 1      |


Output from the biuld:
```
[INFO] --- spotbugs-maven-plugin:4.3.0:check (sca-spotbugs) @ morf-core ---
[INFO] BugInstance size is 25
[INFO] Error size is 0
[INFO] Total bugs: 25
[...]

PR #281, opened 2023-08-16
Upgrade SpotBugs, and allowlist the failing SpotBugs rules

wnicholson:
Continuing #170 work - these changes unblock a move to later JDKs


wnicholson:
I've reversed back to Guice 4.2.2, leaving this as a build-only / SpotBugs-only change.

The Guice upgrade will block a later JDK upgrade but can come with that later.

PR #283, opened 2023-09-13
Switch to SpotBugs

wnicholson:
#170:
- drop FindBugs
- use SpotBugs instead
- bind the plugin to ``verify`` as opposed to ``validate`` (this meant FindBugs wasn't actually doing anything, there was no bytecode to scan)
- allowlist the failing rules


wnicholson:
Thanks @capgen628 👍

Issue #286, opened 2023-10-11
Update build JDK to 17


wnicholson:
[Sonarcloud analysis](https://docs.sonarcloud.io/appendices/scanner-environment/) needs JDK 17 as the build JDK. We should upgrade.

Morf can continue to target Java 8 bytecode (although other work can and should move to targetting 11 IMO).

Morf's dependency on Guice will, I think, need to go up (e.g. to 5.1.0) - this change reversed that update out elsewhere https://github.com/alfasoftware/morf/pull/283/commits/54e75e152d770d09f2d704eaa16cba62dcc7aae0 

Mockito should also probably be upgraded (I believe it's not currently blocking, but is good practice).

wnicholson:
Jacoco 0.8.5 can't instrument, so try moving to 0.8.10.

wnicholson:
Closing this, although it might have caused #296.

PR #287, opened 2023-10-11
Update to build with jdk 17

wnicholson:
#286:
- swap to openjdk17 (but leave the build targetting 8, a later change can move this to 11 which could be major/breaking)
- use Guice 5.1.0, JaCoCo 0.8.10
- change a couple of Javadoc `<h3>` tags that are now errors as (I assume) the upgraded Javadoc tool would prefer to be inside `<h1><h2>`


morf-core/src/main/java/org/alfasoftware/morf/sql/element/FieldReference.java
wnicholson, line 1: Not sure what's going on with the constructor/initialiser changes here, was only expecting to change the javadoc.

jsimlo (approved):

PR #289, opened 2023-10-26
MORF-79 Upgrade Mockito to ver 5.4. Drop support for Java 8.

pstefaniak7:


morf-core/src/test/java/org/alfasoftware/morf/sql/TestObjectTreeTraverserWithSqlElementVisitor.java
```diff
@@ -103,7 +102,6 @@ public void testSimpleJoinCriterionWithSubquery() {
     verify(callback).visit(select2);
     verify(callback).visit(selectsAsField);
     verify(callback).visit(select2.getFields().get(0));
-    verifyNoMoreInteractions(callback);
```
jsimlo, line 106: Is it not important to check?

jsimlo (dismissed):

morf-core/src/test/java/org/alfasoftware/morf/sql/TestObjectTreeTraverserWithSqlElementVisitor.java
pstefaniak7: I think it makes more calls, it wasn't verified properly with the old Mockito

pstefaniak7: so this will fail with new inline mock maker, it is usually more strict

pstefaniak7: @jsimlo Ok, found a better way, see last commit

jsimlo (approved):

gilleain (commented):
Looks good

PR #290, opened 2023-10-27
Add optimistic locking mechanism to the upgrade script to prevent multiple executions

euano:


morf-core/src/main/java/org/alfasoftware/morf/upgrade/UpgradePath.java
```diff
@@ -367,12 +374,27 @@ public UpgradePath create(List<UpgradeStep> steps, ConnectionResources connectio
...
+
+    /**
+     * Creates the initialisation SQL by combining the creation of the upgrade status table, followed by the optimised locking SQL.
+     * The optimised locking statement should always come after, as the locking may be dependent on the upgrade status table.
+     */
+    private List<String> combineInitialisationSql(UpgradeStatusTableService upgradeStatusTableService,
+                                                  List<String> optimisticLockingInitialisationSql) {
+      List<String> updateTableSql = upgradeStatusTableService.updateTableScript(UpgradeStatus.NONE, UpgradeStatus.IN_PROGRESS);
+      updateTableSql.addAll(optimisticLockingInitialisationSql);
+      return updateTableSql;
```
jsimlo, line 397: This relies on `upgradeStatusTableService.updateTableScript()` returning a mutable List, which you can add to.
First, `upgradeStatusTableService.updateTableScript()` shuld return an immutable list, second, this needs to build its own (immutable) list.

morf-core/src/main/java/org/alfasoftware/morf/upgrade/Upgrade.java
```diff
@@ -334,6 +341,35 @@ long getUpgradeAuditRowCount(ResultSetProcessor<Long> processor) {
...
+    SelectStatement selectStatement = select().from(upgradeStatusTable)
+        .where(neq(selectUpgradeAuditTableCount().asField(), upgradeAuditCount))
+        .build();
+
+    InsertStatement insertStatement = insert().into(upgradeStatusTable)
+        .from(selectStatement)
+        .build();
+
+    return connectionResources.sqlDialect().convertStatementToSQL(insertStatement);
+  }
```
jsimlo, line 359: Because all the tests are mocked, we do not actually test your new code on an actual database.
We do not therefore test positive and negative scenarios.

For example, this locking relies on zzzUpgradeStatus having both a primary key and at least one record.
If someone reworks the logic around this table and either removes the primary key or forgets to insert at least one record, your lock will still happily execute, but it will stop locking.

At the minimum, we have to ensure the zzzUpgradeStatus table, created in UpgradeStatusTableServiceImpl.updateTableScript(), is doing both the PK and the insert. But better than that, we should have a test where UpgradeStatusTableService and Upgrade are shown to do the locking, and prove that the lock actually works and allows/stops the upgrade in different scenarios.

At the moment, your locking is intertwined with this Upgrade class, and therefore must be tested within the entire Upgrade logic, which is perhaps not too bad, and we might have such non-mocked test already, which you can use to test this.

Alternatively, extracting this locking into a new class would allow you to do something as simple as TestDatabaseUpgradeLockService, just to make sure that your locking and UpgradeStatusTableService work as expected.

jsimlo (changes requested):

morf-integration-test/src/test/java/org/alfasoftware/morf/upgrade/TestDatabaseUpgradePathValidationService.java
```diff
@@ -0,0 +1,126 @@
...
+    // Add a record to UpgradeAudit and drop the zzzUpgradeStatus table after first run as this is what would happen in an actual upgrade
+    addUpgradeAuditRecord();
+    dropUpgradeStatusTable();
+
+    // Then
+    RuntimeSqlException exception = assertThrows(RuntimeSqlException.class, () -> sqlScriptExecutorProvider.get().execute(initialisationSql));
+    // SQL State 23505 indicates a unique constraint violation on H2/PGSQL and 23000 on Oracle, so match based on either
+    String upgradeStatusUniqueConstraintViolationRegex = "Error executing SQL \\[INSERT INTO [A-Za-z0-9]*[_.]?zzzUpgradeStatus .*? SQL state \\[(23505|23000)]";
+    assertTrue("Should have been a unique constraint violation exception thrown", exception.getMessage().matches(upgradeStatusUniqueConstraintViolationRegex));
+  }
```
jsimlo, line 108: That `SQL state \\[(23505|23000)]` might get different results on different platforms.
We have tested on H2/Postgres/Oracle, which is what we want at the moment.
This might need updating for other platforms in the future, if we start supporting them.

Note: There is a benefit to listing numbers here, we want to check for _unique constraint violation_ explicitly.
The query could fail for other reasons, for example, the table not being present, or disk being full, etc.

jsimlo (approved):

jsimlo:
0.0% Coverage on New Code is an error from the build system. Coverage seems good.

PR #294, opened 2023-11-07
Avoid a double build on Travis

capgen628:


capgen628:
This change will mean we no longer publish SNAPSHOT builds from main. But will verify target.

wnicholson (approved):

PR #299, opened 2023-11-14
Upgrade plugins

harrisric:
Upgrade all standard maven plugins to latest versions.
Fix checkstyle configuration - should be inputEncoding.


harrisric:
Performed a full before/after comparison of the generated artefacts and everything shows up as the same (except timestamps in the generated javadoc files)

jonknight73 (approved):

PR #300, opened 2023-11-20
Ensure optimistic locking doesn't throw an exception when we cannot connect to UpgradeAudit table prior to generating script

euano:


morf-integration-test/src/test/java/org/alfasoftware/morf/upgrade/TestDatabaseUpgradePathValidationService.java
```diff
@@ -103,11 +103,29 @@ public void testInitialisationSql() {
     // Then
     RuntimeSqlException exception = assertThrows(RuntimeSqlException.class, () -> sqlScriptExecutorProvider.get().execute(initialisationSql));
     // SQL State 23505 indicates a unique constraint violation on H2/PGSQL and 23000 on Oracle, so match based on either
-    String upgradeStatusUniqueConstraintViolationRegex = "Error executing SQL \\[INSERT INTO [A-Za-z0-9]*[_.]?zzzUpgradeStatus .*? SQL state \\[(23505|23000)]";
+    String upgradeStatusUniqueConstraintViolationRegex = "Error executing SQL \\[INSERT INTO [A-Za-z0-9_]*[.]?zzzUpgradeStatus .*? SQL state \\[(23505|23000)]";
```
jsimlo, line 106: I like the `[.]?` which has the same meaning as `\.?`
The former is potentially slightly more expensive (memory-wise), but it does not matter within a test.

morf-integration-test/src/test/java/org/alfasoftware/morf/upgrade/TestDatabaseUpgradePathValidationService.java
```diff
@@ -103,11 +103,29 @@ public void testInitialisationSql() {
...
+    // Given
+    List<String> initialisationSql = databaseUpgradePathValidationService.getPathValidationSql(-1); // -1 count returned when unable to connect to UpgradeAudit table
+
+    // When
+    // Execute initialisation SQL for the first time (perform an upgrade)
+    int rowsUpdated = sqlScriptExecutorProvider.get().execute(initialisationSql);
+
+    // Then
+    // No exception being thrown indicates the statement does not insert a duplicate record when the count is -1
+  }
```
jsimlo, line 126: The `int rowsUpdated` can be asserted to be 0.

jsimlo (changes requested):

morf-integration-test/src/test/java/org/alfasoftware/morf/upgrade/TestDatabaseUpgradePathValidationService.java
euano: I did think about doing that but I would be asserting that it was 1, rather than 0, as the initialisation SQL involves inserting a row into zzzUpgradeStatus.  I was thinking by testing that it would make the test more brittle if the statements around the zzzUpgradeStatus table changed but I suppose now that it is coupled to the optimisation SQL, as part of this service, it actually makes sense to test it.  Will update, thanks

morf-integration-test/src/test/java/org/alfasoftware/morf/upgrade/TestDatabaseUpgradePathValidationService.java
euano: Good to know, thanks

jsimlo (approved):

jsimlo:
Note: `0.0% Coverage` is a build error. Coverage is fine.

PR #303, opened 2024-01-09
add regex check on schemaName

ParkinMike:


morf-core/src/main/java/org/alfasoftware/morf/util/SchemaValidatorUtil.java
```diff
@@ -0,0 +1,23 @@
+package org.alfasoftware.morf.util;
+
+/**
+ * Util class to help validate Schema details
+ */
+public class SchemaValidatorUtil {
+
+    private static final String pattern = "^[A-Za-z0-9_]*$";
```
jsimlo, line 8: Rename to `schemaNameValidationPattern`

morf-core/src/main/java/org/alfasoftware/morf/util/SchemaValidatorUtil.java
```diff
@@ -0,0 +1,23 @@
...
+
+    /**
+     * throws a IllegalArgumentException when schemaName contains illegal chars.
+     * @param schemaName - String value of schemaName.
+     * @return - a string containing schemaName.
+     * @throws IllegalArgumentException - when the schemaName contains invalid chars.
+     */
+    public static String validateSchemaName(String schemaName) {
+        if(!schemaName.matches(pattern)) {
+            throw new IllegalArgumentException("schemaName has failed validation check.");
```
jsimlo, line 18: Probably best to include the schemaName in the exception message,
e.g. `"schemaName [" + schemaName + "] failed validation check"`
(also, no dot at the end of the message)

jsimlo (changes requested):

morf-core/src/main/java/org/alfasoftware/morf/util/SchemaValidatorUtil.java
ParkinMike: I thought as there is a chance the user input could be malicious we would not want to log the schemaName? if not I can change this.

jsimlo (approved):

PR #305, opened 2024-01-12
Add result set validation to enforce non-zero source record count

marcellor:
Add compare methods enforcing a non-zero source record count.

### DIFF
--- a/morf-core/src/main/java/org/alfasoftware/morf/jdbc/ResultSetComparer.java
+++ b/morf-core/src/main/java/org/alfasoftware/morf/jdbc/ResultSetComparer.java
@@ -206,7 +206,41 @@ public class ResultSetComparer {
    * @return the number of mismatches between the two data sets.
    */
   public int compare(int[] keyColumns, SelectStatement left, SelectStatement right, Connection connection, CompareCallback callback) {
-    return compare(keyColumns, left, right, connection, connection, callback);
+    return compare(keyColumns, left, right, connection, connection, callback, true);
+  }
+
+
+  /**
+   * Given 2 data sets, return the number of mismatches between them, and
+   * callback with the details of any mismatches as they are found. This method
+   * will generate the result sets itself by executing two select statements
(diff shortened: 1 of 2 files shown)


morf-core/src/main/java/org/alfasoftware/morf/jdbc/ResultSetComparer.java
```diff
@@ -641,4 +702,75 @@ public interface CompareCallback {
...
+       try {
+         if (!rightRs.isBeforeFirst()) {
+           throw new IllegalStateException(format("The following query should return at least one record: [%s]", rightRs.getStatement()));
+         }
+      } catch (SQLException e) {
+        ExceptionUtils.rethrow(e);
+      }
+     }),
+
+     NON_ZERO_RECORD_COUNT_ON_LEFT_AND_RIGHT((leftRs, rightRs) -> {
```
capgen628, line 736: Can we remove this one. It should be equivalent to using both the left and right counterparts.

In fact, since you've already the test case set up it should be easy to prove and increase test coverage all at the same time.

morf-core/src/main/java/org/alfasoftware/morf/jdbc/ResultSetComparer.java
```diff
@@ -144,6 +149,10 @@ public ResultSetComparer create() {
   /** Value used to represent a field value where the record does not exist in the record set */
   public static final String RECORD_NOT_PRESENT = "<Not present>";
 
+  public static final ResultSet leftRs = null;
+
+  public static final ResultSet rightRs = null;
+
```
jsimlo, line 155: I think these are unused

jsimlo (changes requested):

morf-core/src/main/java/org/alfasoftware/morf/jdbc/ResultSetComparer.java
marcellor: Removed.

jsimlo (approved):

PR #307, opened 2024-01-17
Refactor result set validation enums, and add record count query validation

marcellor:
Refactor result set validation enums, and add record count query validation.

### DIFF
--- a/morf-core/src/main/java/org/alfasoftware/morf/jdbc/ResultSetComparer.java
+++ b/morf-core/src/main/java/org/alfasoftware/morf/jdbc/ResultSetComparer.java
@@ -39,6 +39,7 @@ import static org.alfasoftware.morf.jdbc.ResultSetMismatch.MismatchType.MISSING_
 
 import java.math.BigDecimal;
 import java.sql.Connection;
+import java.sql.PreparedStatement;
 import java.sql.ResultSet;
 import java.sql.ResultSetMetaData;
 import java.sql.SQLException;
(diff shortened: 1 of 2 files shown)


morf-core/src/main/java/org/alfasoftware/morf/jdbc/ResultSetComparer.java
```diff
@@ -707,44 +673,59 @@ public interface CompareCallback {
...
+
+    /**
+     * Handles validation of the left and right {@link Comparable} values when a single result is expected.
+     *
+     * @param leftValue the left {@link Comparable} value to validate.
+     * @param rightValue the right {@link Comparable} value to validate.
+     */
+    @SuppressWarnings("unused")
+    void validateSingleResult(Comparable<?> leftValue, Comparable<?> rightValue) { /* Default empty implementation */ }
   }
```
jsimlo, line 729: Name of the `validateBeforeNext()` method feels somewhat confusing to me, because it is not being executed before each "next", which is how I seem to be understanding that name.

Perhaps `validateBeforeFirstRow()` or `validateBeforeComparison()` would be more self-documenting?

All this said, this is an internal part of this class, and thus is not very visible to the outside world, apart from the stacktrace of the validation exception.

morf-core/src/main/java/org/alfasoftware/morf/jdbc/ResultSetComparer.java
```diff
@@ -707,44 +673,59 @@ public interface CompareCallback {
...
+     * @author Copyright (c) Alfa Financial Software Limited. 2024
+     */
+    NON_EMPTY_RESULT {
         @Override
-        void validate(ResultSet leftRs, ResultSet rightRs) {
+        void validateBeforeNext(ResultSet leftRs, ResultSet rightRs) {
           try {
-             if (!leftRs.isBeforeFirst()) {
-                throw new IllegalStateException(format("The following query should return at least one record: [%s]", leftRs.getStatement()));
+             if (!leftRs.isBeforeFirst() || !rightRs.isBeforeFirst()) {
```
capgen628, line 685: Do we need this as an &&?

jsimlo (approved):

PR #310, opened 2024-02-05
Adding SQL tokenizer.

jsimlo:


morf-core/src/main/java/org/alfasoftware/morf/sql/SqlTokenizer.java
```diff
@@ -0,0 +1,707 @@
...
+
+
+  /**
+   * See https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS
+   * See https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/Database-Object-Names-and-Qualifiers.html
+   */
+  private static boolean isIdentifierMidCharacter(char c) {
+    return isIdentifierStartCharacter(c)
+        || '0' <= c && c <= '9'
+      //|| c == '#' // Oracle allows # in identifiers
```
euano, line 623: commented code

morf-core/src/main/java/org/alfasoftware/morf/sql/SqlTokenizer.java
```diff
@@ -0,0 +1,707 @@
...
+  private static boolean isNumberStartCharacter(char c) {
+    return '0' <= c && c <= '9';
+  }
+
+
+  /**
+   * See https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-CONSTANTS-NUMERIC
+   * See https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/Literals.html
+   */
+  private static boolean isNumberMidCharacter(char c) {
```
euano, line 641: is this method necessary?  Could you not just have a single isNumberCharacter() method?

morf-core/src/main/java/org/alfasoftware/morf/sql/SqlTokenizer.java
```diff
@@ -0,0 +1,707 @@
...
+   * <pre>SELECT * FROM table</pre>
+   * <p>Is tokenized as:</p>
+   * <pre>IDENTIFIER(SELECT), WHITESPACE( ), OPERATOR(*), WHITESPACE( ), IDENTIFIER(FROM), WHITESPACE( ), IDENTIFIER(table)</pre>
+   *
+   * @param databaseType ORACLE/PGSQL
+   * @param query Query to be tokenized.
+   * @param visitor Function which will receive TokenType+String pairs.
+   *                The string contains the entire token, as described by the respective TokenType.
+   */
+  public static void tokenizeSqlQuery(String databaseType, String query, Consumer<Token> visitor) {
```
euano, line 137: should we validate the databaseType and throw and exception if it is not Oracle or Postgres, if they're the only supported dialects?

morf-core/src/main/java/org/alfasoftware/morf/sql/SqlTokenizer.java
jsimlo: That actually was intentional. I thought we should just disallow use of # in identifiers if we could.
But then it would have looked like an omission in the code, hence it was left included, but commented out.

jsimlo: I will add a note to the comment making it more obvious

morf-core/src/main/java/org/alfasoftware/morf/sql/SqlTokenizer.java
jsimlo: Depends on whether we will want to start supporting floating numbers like "5.8" in the future.
I had this method in because I thought I would try to read the full numbers, including decimal points, etc.
But then it seemed completely unnecessary for our current purposes, and when reading vendor documentations, dialects seem to differ ever so slightly, which solidified my decision to skip parsing numbers in full detail. Rhis method remained as a remnant of my attempt. It could be useful when we decide to support full numbers in the future. In the meantime, does not do much harm.

morf-core/src/main/java/org/alfasoftware/morf/sql/SqlTokenizer.java
jsimlo: I thought about that, and also the implications of using a null (i.e. unknown dialect).
The thing is that {{NamedParameterPreparedStatement}} can still work without a dialect.
Thus we need to be do our best with a null. Which means we can do our best with unknown dialects too.

If we ever drop the nullability of databaseType, we can perhaps also look at parsing H2 peculiarities.
I have quite obviously overlooked H2 in my implementation.. :)

morf-core/src/main/java/org/alfasoftware/morf/sql/SqlTokenizer.java
```diff
@@ -0,0 +1,707 @@
...
+    final int length = query.length();
+    int i = 0;
+
+    while (i < length) {
+
```
lbriou, line 142: Do you think it would be worth extracting some of this logic outside the loop here?

i = processNextToken(databaseType, query, length, i, visitor);
^ inside the loop

 private int processNextToken(String databaseType, String query, int length, int i, Consumer<Token> visitor) { ... }
^ the same logic extracted outside the loop

jsimlo:
Quality Gate reviewed.

PR #315, opened 2024-03-06
Feature/morf 93 support sequence during upgrades

luketimmons:
Support the addition and removal of schema level sequences during database upgrade steps.


morf-core/src/main/java/org/alfasoftware/morf/jdbc/DatabaseMetaDataProvider.java
```diff
@@ -888,6 +1016,17 @@ protected static AName named(String name) {
...
 
+  /**
+   * Creates a map of all sequence names, indexed by their case-agnostic names. The sequences are retrieved from the
+   * JDBC connection as a result set, where the result set is converted into a map of real sequence names with system
+   * sequences not included.
+   *
+   * @return Map of real sequence names.
+   */
+  public Map<AName, RealName> getSequenceResultSet(String schemaName) {
+    return ImmutableMap.of();
```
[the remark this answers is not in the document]
pgghosh: I think the dialect specific overrides provide the required implementation.

morf-sqlserver/src/main/java/org/alfasoftware/morf/jdbc/sqlserver/SqlServerDialect.java
```diff
@@ -155,6 +157,31 @@ public Collection<String> internalTableDeploymentStatements(Table table) {
...
+  protected Collection<String> internalSequenceDeploymentStatements(Sequence sequence) {
+    List <String> statements = new ArrayList<>();
+
+    StringBuilder createSequenceStatement = new StringBuilder();
+    createSequenceStatement.append("CREATE ");
+
+    createSequenceStatement.append("SEQUENCE ");
+    createSequenceStatement.append(schemaNamePrefix());
+    createSequenceStatement.append(sequence.getName());
+    createSequenceStatement.append(" ");
```
jsimlo, line 173: Same here with the sapce.

morf-sqlserver/src/main/java/org/alfasoftware/morf/jdbc/sqlserver/SqlServerDialect.java
```diff
@@ -479,7 +506,28 @@ protected String getSqlFrom(UpdateStatement statement) {
...
+  @Override
+    protected String getSqlFrom(SequenceReference sequenceReference) {
+    StringBuilder result = new StringBuilder();
+
+    switch (sequenceReference.getTypeOfOperation()) {
+      case NEXT_VALUE:
+        result.append("NEXT VALUE FOR ");
+        break;
+      case CURRENT_VALUE:
+        return "current_value FROM sys.sequences WHERE name = '" + sequenceReference.getName() + "'";
```
jsimlo, line 522: I think this is wrong.
It has to be a full sub-select, including the brackets to make it a field-value again:
```
return "(SELECT current_value FROM sys.sequences WHERE name = '" + sequenceReference.getName() + "')";
````

File: morf-sqlserver/src/main/java/org/alfasoftware/morf/jdbc/sqlserver/SqlServerDialect.java [hunk 4]
```diff
@@ -479,7 +506,28 @@ protected String getSqlFrom(UpdateStatement statement) {
...
+    }
+
+    result.append(sequenceReference.getName());
```
jsimlo, line 525: This does not work with CURRENT_VALUE, only NEXT_VALUE should use it.
Move this directly to the NEXT_VALUE case.
Check what the test will spit out when used.

morf-sqlserver/src/main/java/org/alfasoftware/morf/jdbc/sqlserver/SqlServerMetaDataProvider.java
```diff
@@ -136,4 +142,91 @@ protected ColumnBuilder setAdditionalColumnMetadata(RealName tableName, ColumnBu
...
+        } finally {
+          resultSet.close();
+        }
+      } finally {
+        statement.close();
+      }
+    } catch (SQLException sqle) {
+      throw new RuntimeSqlException("Error running SQL: " + sql, sqle);
+    }
+  }
```
jsimlo, line 230: Same copy-paste

morf-sqlserver/src/main/java/org/alfasoftware/morf/jdbc/sqlserver/SqlServerDialect.java
jsimlo: Oh, I see the return in CURRENT_VALUE instead of a break!
That's okay then, but you got me! :)

morf-sqlserver/src/test/java/org/alfasoftware/morf/jdbc/sqlserver/TestSqlServerDialect.java
```diff
@@ -1408,4 +1425,22 @@ protected String expectedSelectWithExceptAndDbLinkFormer() {
...
+  }
+
+
+  /**
+   * @see AbstractSqlDialectTest#expectedCurrValForSequence()
+   */
+  @Override
+  protected String expectedCurrValForSequence() {
+    return "current_value FROM sys.sequences WHERE name = 'TestSequence'";
+  }
```
jsimlo, line 1445: If this was part of a bigger select, you would see the problem here.

morf-sqlserver/src/test/java/org/alfasoftware/morf/jdbc/sqlserver/TestSqlServerMetaDataProvider.java
```diff
@@ -0,0 +1,146 @@
...
+            when(resultSet.getString(3)).thenReturn("PRIMARY_INDEX_PK");
+          }
+        }
+
+      } else {
+        when(resultSet.getString(1)).thenReturn("Sequence1");
+      }
+      return resultSet;
+    }
+  }
```
jsimlo, line 144: And the same copy-paste tidy-up needed.

jsimlo (changes requested):

morf-core/src/main/java/org/alfasoftware/morf/upgrade/adapt/FilteredSchema.java
```diff
@@ -52,6 +54,37 @@ public boolean apply(Table table) {
...
+    })), Collections2.filter(baseSchema.sequences(), not(new Predicate<Sequence>() {
+      @Override
+      public boolean apply(Sequence sequence) {
+        // String.CASE_INSENSITIVE_ORDER lets you use case-insensitive .contains(Object)
+        Set<String> caseInsensitiveSet = new TreeSet<>(String.CASE_INSENSITIVE_ORDER);
+        Collections.addAll(caseInsensitiveSet, removedSequences);
+        return caseInsensitiveSet.contains(sequence.getName());
+      }
     })));
   }
```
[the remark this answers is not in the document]
luketimmons: I think the issue here would be that we would have two constructors with FilteredSchema(final Schema, final String...) with no way to identify if it were tables or sequences being passed in.

Potentially changing it to FilteredSchema(final Schema, final Sequence...) would be better in this case?

morf-core/src/main/java/org/alfasoftware/morf/metadata/SchemaUtils.java
```diff
@@ -122,7 +133,18 @@ public static Schema schema(Iterable<Table> tables) {
...
+
+  /**
+   * Build a {@link Schema} from a list of {@link Sequence}s.
+   *
+   * @param sequences The sequences to use.
+   * @return A {@link Schema} implementation
+   */
+  public static Schema schema(List<Sequence> sequences) {
+    return new SchemaBean(ImmutableList.of(), ImmutableList.of(), sequences);
   }
```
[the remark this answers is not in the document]
luketimmons: I think that the issue here is that as all of these methods have the same erasure if they are moved over to have Iterables as the argument. So you could have: `schema(Iterable<Table> tables)`, `schema(Iterable<View> views)`, and `schema(Iterable<Sequence> sequences)`. But if you were to call `schema()`, there would be no way to identify which method to use

morf-oracle/src/main/java/org/alfasoftware/morf/jdbc/oracle/OracleMetaDataProvider.java
```diff
@@ -561,6 +580,54 @@ public void handle(ResultSet resultSet) throws SQLException {
...
+   * is slow on Oracle, this uses an optimised query.
+   *
+   * @see <a href="http://docs.oracle.com/cd/B19306_01/server.102/b14237/statviews_2117.htm">ALL_VIEWS specification</a>
+   */
+  private void readSequenceMap() {
+    log.info("Starting read of sequence definitions");
+
+    long start = System.currentTimeMillis();
+
+    final String sequencesSql = "SELECT sequence_name FROM ALL_SEQUENCES WHERE cache_size != 2000 AND sequence_owner=?";
```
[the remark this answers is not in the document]
luketimmons: This was to differentiate between sequences used to create a table's autonumber column and ones that are created independent of tables, i.e. via AddSequence. 

We appear to support adding sequences for tables in Oracle where we default the cache_size to 2000 here: https://github.com/alfasoftware/morf/pull/315/files#diff-216614a0a426b0f31568ca65c0de2d9e692d33c71320c457e280e8612b2f29aeR382

morf-core/src/main/java/org/alfasoftware/morf/metadata/SchemaUtils.java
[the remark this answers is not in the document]
jsimlo: Is that why we use Iterable/Collection/List for each of them? :D
Lolz. Soon we will run out of collection types.

morf-core/src/main/java/org/alfasoftware/morf/upgrade/adapt/FilteredSchema.java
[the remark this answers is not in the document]
jsimlo: Right, this again... ¯\_(ツ)_/¯

morf-oracle/src/main/java/org/alfasoftware/morf/jdbc/oracle/OracleMetaDataProvider.java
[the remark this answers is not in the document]
jsimlo: You might want to say that in a comment, maybe also link to org.alfasoftware.morf.jdbc.oracle.OracleDialect.createNewSequence(Table, Column).
I have completely missed that, others will too.

jsimlo (changes requested):
Two minor changes, all good otherwise.
- comment on OracleMetaDataProvider
- simplification in SequenceBean

morf-core/src/main/java/org/alfasoftware/morf/jdbc/DatabaseMetaDataProvider.java
```diff
@@ -1048,9 +1048,7 @@ protected static AName named(String name) {
    * @param schemaName
    * @return
    */
-  protected String buildSequenceSql(String schemaName) {
-    return null;
-  }
```
jsimlo: Note: I like this change, but it also makes the class abstract, which is a backwards-incompatible change.
That said, there were other changes already here which were backwards-incompatible too, although in less visible places.

jsimlo (approved):

jsimlo:
I have reviewed the remaining [16 New issues](https://sonarcloud.io/project/issues?id=org.alfasoftware%3Amorf-parent&pullRequest=315&resolved=false&inNewCodePeriod=true) from Sonar, they are all fine.

PR #329, opened 2024-09-18
Added portable functions

Amarrd:
These can be used to declare functions which are not currently supported by morf (and with no plan for support).


morf-testsupport/src/main/java/org/alfasoftware/morf/jdbc/AbstractSqlDialectTest.java
```diff
@@ -5064,6 +5065,45 @@ public void testPrepareStatementParameter() throws SQLException {
...
+            "TRANSLATE",
+            new FieldReference("field"),
+            new FieldLiteral("5"),
+            new FieldLiteral("E"))
+        .as("field");
+
+    UpdateStatement testStatement = UpdateStatement.update(new TableReference("Table")).set(function).build();
+
+    assertEquals(expectedPortableStatement(), testDialect.convertStatementToSQL(testStatement));
+  }
```
jsimlo, line 5105: Function name for every dialect is the same, "TRANSLATE". Therefore, technically,
you are not testing the correct choice of function name by dialect, only the arguments differ.
The implementation keeps function names and arguments inseparable, but you should not count on that.

morf-core/src/main/java/org/alfasoftware/morf/sql/element/PortableFunction.java
```diff
@@ -0,0 +1,114 @@
...
+   * @param arguments the arguments for the function
+   * @return the portable function to be used
+   */
+  public static PortableFunction withFunctionForDatabaseType(String databaseTypeIdentifier, String functionName, AliasedField... arguments) {
+    Pair<String, List<AliasedField>> functionWithArguments = Pair.of(functionName, Arrays.asList(arguments));
+    Map<String, Pair<String, List<AliasedField>>> databaseFunctionMap = new HashMap<>();
+
+    databaseFunctionMap.put(databaseTypeIdentifier, functionWithArguments);
+    return new PortableFunction(databaseFunctionMap);
+  }
```
jsimlo, line 56: I'd prefer a more comfortable builder pattern here.

jsimlo (changes requested):

morf-core/src/main/java/org/alfasoftware/morf/sql/element/PortableFunction.java
Amarrd: Updated approach to use builder pattern

morf-testsupport/src/main/java/org/alfasoftware/morf/jdbc/AbstractSqlDialectTest.java
Amarrd: Tweaked test coverate so that this is tested in AbstractSqlDialectTest. I was previously doing this testing in TestPortableSqlFunction, but it makes more sense here.

morf-core/src/main/java/org/alfasoftware/morf/sql/element/PortableSqlFunction.java
```diff
@@ -0,0 +1,133 @@
...
+            Pair<String, List<AliasedField>> functionWithArguments = Pair.of(functionName, Arrays.asList(arguments));
+            databaseFunctionMap.put(databaseTypeIdentifier, functionWithArguments);
+
+            return this;
+        }
+
+        public Builder as(String alias) {
+            this.alias = alias;
+            return this;
+        }
```
jsimlo, line 73: You should also get the alias via `PortableSqlFunction.builder().....build().as("alias")`
If that works, we probably do not even need this one?

morf-testsupport/src/main/java/org/alfasoftware/morf/jdbc/AbstractSqlDialectTest.java
```diff
@@ -5067,37 +5067,39 @@ public void testPrepareStatementParameter() throws SQLException {
...
+                    new FieldLiteral("4"),
+                    new FieldLiteral("D"))
+            .withFunctionForDatabaseType(
+                    "SQL_SERVER",
+                    "SOUNDEX",
+                    new FieldReference("field"),
+                    new FieldLiteral("5"),
+                    new FieldLiteral("E"))
+            .as("field")
+            .build();
```
jsimlo, line 5102: You should also get the alias via `PortableSqlFunction.builder().....build().as("alias")`
Testing that path is probably slightly more important, because that is what users would use out of habit.

jsimlo (changes requested):

morf-core/src/main/java/org/alfasoftware/morf/sql/element/PortableSqlFunction.java
Amarrd: Following the pattern of implementing AliasedFieldBuilder

morf-testsupport/src/main/java/org/alfasoftware/morf/jdbc/AbstractSqlDialectTest.java
Amarrd: Following the pattern of implementing AliasedFieldBuilder

jsimlo (approved):

PR #330, opened 2024-10-08
Adding SchemaAutoHealer

jsimlo:
Notes:
- This new SchemaAutoHealer is for auto-healing the source Schema, as read from the database. It is dialect agnostic and can modify the source Schema.
- The older SqlDialect.getSchemaConsistencyStatements() is for auto-healing hidden inconsistencies within the database, which are invisible via the source Schema.

### DIFF
--- a/morf-core/src/main/java/org/alfasoftware/morf/jdbc/SqlDialect.java
+++ b/morf-core/src/main/java/org/alfasoftware/morf/jdbc/SqlDialect.java
@@ -4457,7 +4457,7 @@ public abstract class SqlDialect {
   /**
    * Returns any statements needed to automatically heal the given schema.
    *
-   * This healer is intended for automated database modifications transparent to the {@link Schema}.
+   * This healer is intended for automated database modifications not visible via the {@link Schema}.
    * For example the names of primary key indexes are not generally available via the {@link Schema},
    * and can therefore be healed via this method without disrupting the {@link Schema} contents.
    *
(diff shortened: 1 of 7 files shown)


morf-core/src/main/java/org/alfasoftware/morf/jdbc/SqlDialect.java
```diff
@@ -4456,6 +4457,19 @@ protected String getSqlForPortableFunction(Pair<String, List<AliasedField>> func
   /**
    * Returns any statements needed to automatically heal the given schema.
    *
+   * This healer is intended for automated database modifications transparent to the {@link Schema}.
```
euano, line 4460: I think this statement contradicts the rest of the javadoc?  Specifically 'modifications transparent to the Schema', my understanding based on reading the rest of the javadoc is that the method is intended for modifications to things not visible by the Schema?

morf-core/src/main/java/org/alfasoftware/morf/upgrade/SchemaAutoHealer.java
```diff
@@ -0,0 +1,120 @@
...
+ * not visible in the {@link Schema}.
+ *
+ * Also note that this type of healer is dialect-agnostic, all dialects will get equally affected.
+ */
+public interface SchemaAutoHealer {
+
+  /**
+   * Analyses given database schema,
+   * and produces a set of statements to be run on the database
+   * to arrive at a healed version of the that schema.
```
euano, line 33: the that schema

euano (approved):

morf-core/src/main/java/org/alfasoftware/morf/jdbc/SqlDialect.java
jsimlo: "modifications to things not visible by the Schema" == "modifications transparent to the Schema"
Is that not the same?

euano: Ohhh ok I see where you're coming from now, transparent as in see through i.e. invisible?  I was reading it more as transparent as in 'it's clear for it to see'.  Either way, it's perhaps more confusing than saying 'modifications not visible via the Schema interface'.  Or perhaps I am just being dumb... I did work out what the method could be sued from the rest of the javadoc anyway, so I suppose it doesn't matter too much

euano (dismissed):

euano (approved):

PR #331, opened 2024-11-13
Adding validation preventing standard_conforming_strings=off

jsimlo:


morf-core/src/test/java/org/alfasoftware/morf/jdbc/TestAbstractConnectionResources.java
euano, line 1: Javadoc or a comment for some of the tests (the happy-path ones) might make it a bit clearer exactly what behaviour they're testing, but having looked at the changes in the source code I can see what the intention is (that no exception is thrown).

jsimlo: Good call, I have added javadocs

euano (approved):

PR #333, opened 2024-12-04
Feature/ignore partition tables postgres

BrunoSMonteiro74:
- Add partitions as ignore tables
- Add partitioned tables to table list as regular tables


morf-postgresql/src/main/java/org/alfasoftware/morf/jdbc/postgresql/PostgreSQLMetaDataProvider.java
```diff
@@ -51,6 +53,47 @@ public PostgreSQLMetaDataProvider(Connection connection, String schemaName) {
...
+    try(Statement ignoredTablesStmt = connection.createStatement()) {
+      // distinguish partitioned tables from regular ones: relkind = 'p' (partition) or 'r' (regular) also can use boolean col relispartition
+      // a partition table attached has (r, true)
+      try (ResultSet ignoredTablesRs = ignoredTablesStmt.executeQuery("select relname from pg_class where relispartition and relkind = 'r'")) {
+        while (ignoredTablesRs.next()) {
+          ignoredTables.add(ignoredTablesRs.getString(1).toLowerCase(Locale.ROOT));
+        }
+      }
+    } catch (SQLException e) {
+        // ignore exception, if it fails then incompatible Postgres version
```
jsimlo, line 69: This could be a problem. Connections where statements fail are usually marked as rollback-only on Postgres.

morf-postgresql/src/main/java/org/alfasoftware/morf/jdbc/postgresql/PostgreSQLMetaDataProvider.java
```diff
@@ -51,6 +53,47 @@ public PostgreSQLMetaDataProvider(Connection connection, String schemaName) {
...
+      }
+    } catch (SQLException e) {
+      // ignore exception, if it fails then incompatible Postgres version
+    }
+    return partitionedTables;
+  }
+
+
+  @Override
+  protected boolean isIgnoredTable(@SuppressWarnings("unused") RealName tableName) {
```
jsimlo: `@SuppressWarnings("unused")`

morf-postgresql/src/main/java/org/alfasoftware/morf/jdbc/postgresql/PostgreSQLMetaDataProvider.java
```diff
@@ -51,6 +53,47 @@ public PostgreSQLMetaDataProvider(Connection connection, String schemaName) {
...
+    return ignoredTables.get().contains(tableName.getDbName().toLowerCase(Locale.ROOT));
```
jsimlo: Should also call or super.isIgnoredTable(tableName)

morf-core/src/main/java/org/alfasoftware/morf/jdbc/DatabaseMetaDataProvider.java
```diff
@@ -116,6 +118,8 @@ public abstract class DatabaseMetaDataProvider implements Schema {
   private final LoadingCache<AName, Sequence> sequenceCache = CacheBuilder.newBuilder().build(CacheLoader.from(this::loadSequence));
 
   private final Supplier<Map<String, String>> databaseInformation = Suppliers.memoize(this::loadDatabaseInformation);
+  protected Supplier<Set<String>> ignoredTables = Suppliers.memoize(this::getIgnoredTables);
+  protected Supplier<Set<String>> partitionedTables = Suppliers.memoize(this::getPartitionedTables);
```
jsimlo: You should call these loadIgnoredTables and loadPartitionedTables, everything else is named this way

morf-postgresql/src/main/java/org/alfasoftware/morf/jdbc/postgresql/PostgreSQLMetaDataProvider.java
```diff
@@ -51,6 +53,47 @@ public PostgreSQLMetaDataProvider(Connection connection, String schemaName) {
...
+      // a partition table attached has (r, true)
+      try (ResultSet ignoredTablesRs = ignoredTablesStmt.executeQuery("select relname from pg_class where relispartition and relkind = 'r'")) {
+        while (ignoredTablesRs.next()) {
+          ignoredTables.add(ignoredTablesRs.getString(1).toLowerCase(Locale.ROOT));
+        }
+      }
+    } catch (SQLException e) {
+        // ignore exception, if it fails then incompatible Postgres version
+    }
+    return ignoredTables;
```
jsimlo: Use ImmutableSet.Builder for ignoredTables to ensure the result of the loading is immutable.

morf-postgresql/src/main/java/org/alfasoftware/morf/jdbc/postgresql/PostgreSQLMetaDataProvider.java
```diff
@@ -51,6 +53,47 @@ public PostgreSQLMetaDataProvider(Connection connection, String schemaName) {
...
+      // a partition table has (p, false)
+      try (ResultSet ignoredTablesRs = partitionedTablesStmt.executeQuery("select relname from pg_class where not relispartition and relkind = 'p'")) {
+        while (ignoredTablesRs.next()) {
+          partitionedTables.add(ignoredTablesRs.getString(1).toLowerCase(Locale.ROOT));
+        }
+      }
+    } catch (SQLException e) {
+      // ignore exception, if it fails then incompatible Postgres version
+    }
+    return partitionedTables;
```
jsimlo: Use ImmutableSet.Builder for partitionedTables to ensure the result of the loading is immutable.

morf-postgresql/src/main/java/org/alfasoftware/morf/jdbc/postgresql/PostgreSQLMetaDataProvider.java
```diff
@@ -51,6 +53,47 @@ public PostgreSQLMetaDataProvider(Connection connection, String schemaName) {
...
+      // distinguish partitioned tables from regular ones: relkind = 'p' (partition) or 'r' (regular) also can use boolean col relispartition
+      // a partition table attached has (r, true)
+      // a partition table has (p, false)
+      try (ResultSet ignoredTablesRs = partitionedTablesStmt.executeQuery("select relname from pg_class where not relispartition and relkind = 'p'")) {
+        while (ignoredTablesRs.next()) {
+          partitionedTables.add(ignoredTablesRs.getString(1).toLowerCase(Locale.ROOT));
+        }
+      }
+    } catch (SQLException e) {
+      // ignore exception, if it fails then incompatible Postgres version
```
jsimlo, line 87: Could this be a problem? Connections where statements fail are usually marked as rollback-only on Postgres.

jsimlo (changes requested):

morf-postgresql/src/main/java/org/alfasoftware/morf/jdbc/postgresql/PostgreSQLMetaDataProvider.java
BrunoSMonteiro74: I don't expect to get an incompatible postgres version. The partition support for list and range has been around since 2017 version 10. The hash partition was added on version 11.

morf-postgresql/src/main/java/org/alfasoftware/morf/jdbc/postgresql/PostgreSQLMetaDataProvider.java
BrunoSMonteiro74: I don't expect to get an incompatible postgres version. The partition support for list and range has been around since 2017 version 10. The hash partition was added on version 11.

morf-postgresql/src/main/java/org/alfasoftware/morf/jdbc/postgresql/PostgreSQLMetaDataProvider.java
jsimlo: At least add a log line :)

morf-integration-test/src/test/java/org/alfasoftware/morf/upgrade/TestDatabaseUpgradePathValidationService.java
```diff
@@ -82,6 +82,8 @@ public void setup() {
   public void tearDown() {
     dropUpgradeStatusTable();
     schemaManager.invalidateCache();
+    // to make following test on test suite run clean - org.alfasoftware.morf.upgrade.TestFullDeployment
+    schemaManager.dropAllTables();
```
jsimlo, line 86: Agree with the change, but the comment is unnecessary, since the order of tests will change in the future. It will be other tests that benefit from properly clearing down the database.

morf-core/src/main/java/org/alfasoftware/morf/metadata/AdditionalMetadata.java
```diff
@@ -18,6 +19,26 @@ default Map<String, String> primaryKeyIndexNames() {
...
+
+  /**
+   * Provides the names of all partition tables in the database. This applies for now for postgres. Note that the order of
+   * the tables in the result is not specified. The case of the
+   * table names may be preserved when logging progress, but should not be relied on for schema
+   * processing. A partition table is a table that is a partition of a partitioned table.
+   *
+   * @return A collection of all partition table names available in the database.
+   */
+  default Collection<String> partitionTableNames() { throw new NotImplementedException("Not implemented yet."); }
```
jsimlo, line 40: Not particularly important, but if you make further changes, change both of these to ```Set<String>```.

morf-postgresql/src/main/java/org/alfasoftware/morf/jdbc/postgresql/PostgreSQLMetaDataProvider.java
```diff
@@ -51,6 +54,47 @@ public PostgreSQLMetaDataProvider(Connection connection, String schemaName) {
...
+    }
+    return partitionedTables.build();
+  }
+
+
+  @Override
+  protected boolean isIgnoredTable(RealName tableName) {
+    return ignoredTables.get().contains(tableName.getDbName().toLowerCase(Locale.ROOT)) || super.isIgnoredTable(tableName);
+  }
```
jsimlo, line 96: You have added ```ignoredTables``` to ```DatabaseMetaDataProvider``` which is fine.
But this implementation should also go to ```DatabaseMetaDataProvider``` now.
Otherwise, all sorts of confusing bugs could follow in the future.

morf-postgresql/src/main/java/org/alfasoftware/morf/jdbc/postgresql/PostgreSQLMetaDataProvider.java
jsimlo: As above, at least add a log line :)

morf-core/src/main/java/org/alfasoftware/morf/jdbc/DatabaseMetaDataProvider.java
```diff
@@ -116,6 +117,8 @@ public abstract class DatabaseMetaDataProvider implements Schema {
   private final LoadingCache<AName, Sequence> sequenceCache = CacheBuilder.newBuilder().build(CacheLoader.from(this::loadSequence));
 
   private final Supplier<Map<String, String>> databaseInformation = Suppliers.memoize(this::loadDatabaseInformation);
+  protected Supplier<ImmutableSet<String>> ignoredTables = Suppliers.memoize(this::loadIgnoredTables);
+  protected Supplier<ImmutableSet<String>> partitionedTables = Suppliers.memoize(this::loadPartitionedTables);
```
jsimlo, line 121: Change these two to ```Supplier<Set<RealName>>``` (or ```Supplier<Set<AName>>``` if not RealName).
Both for consistency and also to deal with case-sensitivity problem (see my comment about moving ```isIgnoredTable()``` below).

morf-core/src/main/java/org/alfasoftware/morf/jdbc/DatabaseMetaDataProvider.java
```diff
@@ -148,6 +151,9 @@ private Map<String, String> loadDatabaseInformation() {
     }
   }
 
+  protected ImmutableSet<String> loadIgnoredTables() { return ImmutableSet.of(); }
+
+  protected ImmutableSet<String> loadPartitionedTables() { return ImmutableSet.of(); }
```
jsimlo, line 156: Change the return type of these to ```Set```.
Return ```ImmutableSet.of()``` for immutability, but always declare the intended type, e.g. ```Set```.

morf-core/src/main/java/org/alfasoftware/morf/jdbc/DatabaseMetaDataProvider.java
```diff
@@ -306,6 +312,11 @@ protected Map<AName, RealName> loadAllTableNames() {
             throw new RuntimeSqlException("Error reading metadata for table ["+tableName+"]", e);
           }
         }
+        // add partitioned tables to list
+        partitionedTables.get().forEach(table -> {
+          RealName partionedTableName = createRealName(table, table);
+          tableNameMappings.put(partionedTableName, partionedTableName);
+        });
```
jsimlo, line 319: I was wondering, these get ignored via ```isIgnoredTable()``` and therefore then have to be added back? Why?

If we add support for Oracle's partitioning, this might become problematic, unless Oracle also excludes those tables during the above and will need us to also add those tables here.

morf-postgresql/src/main/java/org/alfasoftware/morf/jdbc/postgresql/PostgreSQLMetaDataProvider.java
```diff
@@ -51,6 +54,47 @@ public PostgreSQLMetaDataProvider(Connection connection, String schemaName) {
   }
 
 
+  @Override
+  protected ImmutableSet<String> loadIgnoredTables() {
+    ImmutableSet.Builder<String> ignoredTables = new ImmutableSet.Builder<>();
+    try(Statement ignoredTablesStmt = connection.createStatement()) {
+      // distinguish partitioned tables from regular ones: relkind = 'p' (partition) or 'r' (regular) also can use boolean col relispartition
+      // a partition table attached has (r, true)
```
jsimlo, line 62: Below, you've got the full explanation or partitions and partitioned tables:
```
      // a partition table attached has (r, true)
      // a partition table has (p, false)
```

Add that same here as well, as this thing is kind of confusing at first.

morf-postgresql/src/main/java/org/alfasoftware/morf/jdbc/postgresql/PostgreSQLMetaDataProvider.java
```diff
@@ -233,4 +277,15 @@ protected String buildSequenceSql(String schemaName) {
...
+
+
+  @Override
+  public Collection<String> partitionedTableNames() {
+    return super.partitionedTables.get();
+  }
+
+  @Override
+  public Collection<String> partitionTableNames() {
+    return super.ignoredTables.get();
```
jsimlo, line 289: Well, as soon as someone adds something else to ```ignoredTables```, this will become wrong.
It is quite possible someone will add something  to ```ignoredTables``` in the future.

jsimlo (changes requested):

morf-core/src/main/java/org/alfasoftware/morf/jdbc/DatabaseMetaDataProvider.java
BrunoSMonteiro74: The tables are not added back. We have two types of tables related to partitions in postgres: The Partitioned table MEASURE(id, a, b) PARTITION by ID  -- not ignored, need to be added because the default java API for reading tables in Postgres driver ignores them
and the partition tables 
MEASURE_P1 (id, a, b) FOR VALUES ('0) TO ('37000') -- ignored, the partitions themselves.

morf-postgresql/src/main/java/org/alfasoftware/morf/jdbc/postgresql/PostgreSQLMetaDataProvider.java
BrunoSMonteiro74: Renamed the ignoredTables collection to be specifically for partions. It will not be an encompassing "ignored tables". The partition tables (the ones attached to the partition table) will always be ignored for postgres

morf-postgresql/src/main/java/org/alfasoftware/morf/jdbc/postgresql/PostgreSQLMetaDataProvider.java
BrunoSMonteiro74: improved the comment

morf-core/src/main/java/org/alfasoftware/morf/jdbc/DatabaseMetaDataProvider.java
BrunoSMonteiro74: Changed to Set<RealName>

morf-core/src/main/java/org/alfasoftware/morf/jdbc/DatabaseMetaDataProvider.java
BrunoSMonteiro74: Changed. It was not done initially done because the RealName class is not public. Made it public.

morf-postgresql/src/main/java/org/alfasoftware/morf/jdbc/postgresql/PostgreSQLMetaDataProvider.java
BrunoSMonteiro74: moved

morf-postgresql/src/main/java/org/alfasoftware/morf/jdbc/postgresql/PostgreSQLMetaDataProvider.java
BrunoSMonteiro74: this comment no longer makes sense.

morf-postgresql/src/main/java/org/alfasoftware/morf/jdbc/postgresql/PostgreSQLMetaDataProvider.java
BrunoSMonteiro74: added log line

PR #334, opened 2025-01-15
Concurrent dataset connector

pstefaniak7:


morf-core/src/main/java/org/alfasoftware/morf/dataset/ConcurrentDataSetConnector.java
```diff
@@ -0,0 +1,285 @@
...
+
+
+  /**
+   * Creates a new instance of this class.
+   *
+   * @param producerSupplier The supplier for data to transmit.
+   * @param consumerSupplier The supplier for the target to which the data should be sent.
+   * @param threadCount Uses provided number of threads instead of default
+   */
+  public ConcurrentDataSetConnector(Supplier<DataSetProducer> producerSupplier, Supplier<DataSetConsumer> consumerSupplier, int threadCount) {
```
pstefaniak7, line 84: I was thinking that maybe we could pass boolean progressLogging parameter here to turn it on/off ?

morf-core/src/main/java/org/alfasoftware/morf/jdbc/ConcurrentSchemaModificationAdapter.java
```diff
@@ -0,0 +1,272 @@
+/* Copyright 2017 Alfa Financial Software
+ *
```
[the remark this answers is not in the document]
pstefaniak7: This is fixed now.

morf-core/src/main/java/org/alfasoftware/morf/jdbc/ConcurrentSchemaModificationAdapter.java
```diff
@@ -0,0 +1,271 @@
...
+
+
+  /**
+   * @see org.alfasoftware.morf.dataset.DataSetAdapter#open()
+   */
+  @Override
+  public void open() {
+    schemaLock.lock();
+    try {
+      if(schemaState == null) { //initialize only once
```
pstefaniak7, line 94: This is protected by a lock so code smell complaint is not entirely valid.

morf-core/src/main/java/org/alfasoftware/morf/dataset/ConcurrentDataSetConnector.java
```diff
@@ -0,0 +1,287 @@
...
+      this.totalTableCount = totalTableCount;
+    }
+
+
+    @Override
+    public void run() {
+      while(shouldContinue) {
+        logInternal();
+        try {
+          Thread.sleep(2000);
```
euano, line 281: Was wondering whether every 2 seconds might be a little frequent and result in a serious amount of logs when processing a larger upload such as a 23k agreement subsetting extract?  Think Ryan said he'd seen upload time of ~30mins for a 10k extract.  Granted, this dev should greatly improve those times, but even if it was reduced by 80% that's still 180 log lines for 6 mins processing.  I think just something to show that it's progressing is the idea, maybe every 15/20 seconds or so would be sufficient?

morf-core/src/main/java/org/alfasoftware/morf/dataset/ConcurrentDataSetConnector.java
```diff
@@ -0,0 +1,287 @@
...
+ *
+ * <p>
+ * This version of the connector works using an {@link ExecutorService} to provide
+ * work management and concurrency. It is expecting a thread safe implementation of {@link DataSetProducer}
+ * and a supplier of consumer instances to be executed in parallel.
+ * </p>
+ *
+ */
+
+public class ConcurrentDataSetConnector {
```
euano, line 45: Nice

morf-core/src/main/java/org/alfasoftware/morf/jdbc/ConcurrentSchemaModificationAdapter.java
```diff
@@ -0,0 +1,271 @@
...
+    //this will open the consumer
+    super.open();
+  }
+
+
+  /**
+   * Drops all views from the existing schema if it has not already done so. This should be called whenever tables are dropped or modified to guard against an invalid situation.
+   */
+  private void dropExistingViewsIfNecessary(Connection conn) {
+    if (schemaState.viewsDropped) {
```
euano, line 122: should this access of the schemaState not also be protected by the lock?

euano: or is that what you mean by the code smell complaint not being entirely valid above, it's not actually necessary to lock it for all accesses?

morf-core/src/main/java/org/alfasoftware/morf/dataset/ConcurrentDataSetConnector.java
pstefaniak7: Hmmm, it takes about 20 sec with 4 threads and standard start position. Would 5 or 10 sec log interval be acceptable ?

morf-core/src/main/java/org/alfasoftware/morf/jdbc/ConcurrentSchemaModificationAdapter.java
pstefaniak7: This check should be fine as viewsDropped was made a volatile boolean. Once views are dropped, we exit the function entirely and the lock is not necessary.

morf-core/src/main/java/org/alfasoftware/morf/dataset/ConcurrentDataSetConnector.java
jonknight73: Given that there is debate about the correct value to use, it suggests that it should be configurable - so that it can be overridden downstream to an appropriate value for that context.

euano (dismissed):

euano (approved):

PR #336, opened 2025-01-27
Code to allow renaming of a column that is used in an index

cnaant:


morf-core/src/test/java/org/alfasoftware/morf/upgrade/TestChangeColumn.java
jsimlo, line 1: The changes are fine here; but what has happened to the "whitepaces" or perhaps "newlines" in this file?

jsimlo (approved):

jsimlo (approved):

PR #339, opened 2025-03-19
Improve the feedback provided by UpgradeTestHelper.

mattbattison:
Previously, UpgradeTestHelper would fail immediately when it encountered an invalid upgrade step. This meant that a test using the helper would have to be run multiple times to flush out all the errors.

This change enhances the helper to combine all the errors into a single failure message where possible:
- All public upgrade steps that should be package-visible will be shown as a single failure.
- All errors relating to the sequence, UUID, Jira ID, and description will be shown as a single failure.

### DIFF
--- a/morf-testsupport/src/main/java/org/alfasoftware/morf/testing/UpgradeTestHelper.java
+++ b/morf-testsupport/src/main/java/org/alfasoftware/morf/testing/UpgradeTestHelper.java
@@ -15,19 +15,18 @@
 
 package org.alfasoftware.morf.testing;
 
-import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNotNull;
-import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
 
 import java.lang.reflect.Constructor;
 import java.lang.reflect.Modifier;
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashSet;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Set;
-import java.util.stream.Collectors;
+import java.util.TreeSet;
 
 import org.alfasoftware.morf.jdbc.ConnectionResources;
 import org.alfasoftware.morf.jdbc.DatabaseDataSetProducer;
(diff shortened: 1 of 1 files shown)


morf-testsupport/src/main/java/org/alfasoftware/morf/testing/UpgradeTestHelper.java
```diff
@@ -188,42 +197,63 @@ public void validateUpgradeStepProperties(Iterable<Class<? extends UpgradeStep>>
...
+      errors.add(String.format("Upgrade step [%s] must have a JIRA ID", upgradeStepClass.getSimpleName()));
+    } else {
+      for (String jiraId : StringUtils.split(upgradeStep.getJiraId(), ',')) {
+        if (!jiraIdIsValid(jiraId)) {
+          errors.add(String.format("Upgrade step [%s] must have a valid JIRA ID [%s]", upgradeStepClass.getSimpleName(), upgradeStep.getJiraId()));
+        }
+      }
     }
+
+    return errors;
```
jsimlo, line 256: This returns a mutable ArrayList.
Maybe use ImmutableList instead.
Or wrap this in Collections.unmodifiableList(list)

morf-testsupport/src/main/java/org/alfasoftware/morf/testing/UpgradeTestHelper.java
```diff
@@ -244,22 +274,30 @@ private boolean jiraIdIsValid(final String jiraId) {
...
+
+      errors.addAll(validateUpgradeStep(upgradeStep));
+      upgradeSteps.add(upgradeStep);
+    }
+
+    if (!errors.isEmpty()) {
+      fail(String.format("Upgrade step errors were found:%n%s%n", String.join(System.lineSeparator(), errors)));
+    }
+
+    return upgradeSteps;
```
jsimlo, line 301: This returns a mutable ArrayList.
Maybe use ImmutableList instead.
Or wrap this in Collections.unmodifiableList(list)

jsimlo (changes requested):

jsimlo (approved):

PR #347, opened 2025-07-18
Improve ResultSetComparer

harrisric:
BigDecimal.stripTrailingZeros is inefficient even when there is no possible work to do (at least through J17), so shortcut this.

Minor correction to javadoc and improvement to tablereference aliasing.

### DIFF
--- a/morf-core/src/main/java/org/alfasoftware/morf/sql/element/TableReference.java
+++ b/morf-core/src/main/java/org/alfasoftware/morf/sql/element/TableReference.java
@@ -139,10 +139,10 @@ public class TableReference implements DeepCopyableWithTransformation<TableRefer
    * Specifies the alias to use for the table.
    *
    * @param aliasName the name of the alias
-   * @return an updated {@link TableReference} (this will not be a new object)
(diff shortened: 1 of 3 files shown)


morf-core/src/main/java/org/alfasoftware/morf/sql/element/TableReference.java
```diff
@@ -139,10 +139,10 @@ public TableReference(String tableName, boolean temporary) {
    * Specifies the alias to use for the table.
    *
    * @param aliasName the name of the alias
-   * @return an updated {@link TableReference} (this will not be a new object)
+   * @return an updated {@link TableReference} (this will not be a new object unless using immutable DSL option and the alias is being changed)
    */
   public TableReference as(String aliasName) {
-    if (AliasedField.immutableDslEnabled()) {
+    if (AliasedField.immutableDslEnabled() && !aliasName.equals(alias)) {
```
jsimlo, line 145: Good change.
Generally, it could have lead to problems (usage relying on new object being created here),
however, this is the only mutating method on this class right now, and therefore no problems possible.

jsimlo (approved):

PR #371, opened 2026-03-06
initial development of entity based changelogs

CaIebBird:
Current changelogs are based by version number, rather than entity.
This development enables the use of entity based changelogs in parallel.


CaIebBird:
Hello @BrannanH, can you please review the code changes?

PR #375, opened 2026-04-20
DeployedIndexes: unified tracking table for all indexes (full model)

michal-morzywolek:
## Summary

Introduces a `DeployedIndexes` tracking table that records **every index** managed by Morf (deferred and non-deferred), along with the lifecycle of any deferred index (`PENDING` → `IN_PROGRESS` → `COMPLETED`/`FAILED`).

- Declare a deferred index via `index("name").deferred().columns(...)`.
- After upgrade, adopters call `UpgradePath.getDeferredIndexStatements()` to fetch `DeferredIndexJob` objects (table + index + SQL) and execute them at their own pace.
- `DeployedIndexTracker` exposes `markStarted / markCompleted / markFailed` for status reporting.
- Enricher merges physical schema with tracking rows at upgrade start and performs consistency checks (untracked physical index, tracked-but-missing non-deferred, orphan rows).
- Dialect matrix: PostgreSQL `CONCURRENTLY`, Oracle `ONLINE PARALLEL NOLOGGING`, H2 standard CREATE INDEX; MySQL / SQL Server silently ignore `.deferred()` (no background creation support).

This branch is the **full tracking** alternative. See the sibling PR for the **slim** (deferred-only) variant — an A/B comparison.

## Test plan
- [ ] Review `TestDeployedIndexesIntegration` — full lifecycle coverage on H2.
- [ ] Review enricher consistency-check tests (`TestDeployedIndexesModelEnricherImpl`).
- [ ] Confirm `supportsDeferredIndexCreation()` returning `false` on MySQL / SQL Server falls back to immediate `CREATE INDEX` with a `COMPLETED` tracking row.
- [ ] Manual: run an upgrade, assert `getDeferredIndexStatements()` returns the expected `DeferredIndexJob`s, execute them, verify tracker transitions.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #376, opened 2026-04-20
DeployedIndexes slim: deferred-only tracking (A/B alternative to #375)

michal-morzywolek:
## Summary

Slim redesign of the `DeployedIndexes` tracking table: rows exist **only for deferred indexes**. Non-deferred indexes live only in the physical DB and are policed by the framework's existing `SchemaHomology.schemasMatch` comparison (same machinery that validates upgrade paths).

The full-tracking alternative is #375. Same adopter API, same per-database semantics, same configuration — just a narrower scope and a lot less code.

### Why this is simpler

1. **Scope drop eliminates redundant work.** Morf already compares declared vs. physical schema before every upgrade — that's how paths are validated. The full model re-implemented variants of that check inside its own enricher; slim just uses the existing authoritative machinery (~200 lines gone).
2. **No prepopulation → no new upgrade-step API.** The full model had to seed a tracking row for every pre-existing index, which forced a new 3-arg `UpgradeStep.execute` overload plus a supporting context interface so the seeding step could read pre-upgrade schema. Slim has nothing to seed → the whole side-API is deleted.
3. **The build lifecycle — the table's real job — is fully preserved.** `PENDING` → `IN_PROGRESS` → `COMPLETED`/`FAILED` is the one state schema comparison can't see (a declared-deferred index legitimately isn't physical yet). Slim keeps all of it, including cross-upgrade carry-forward of unbuilt deferred indexes.
4. **Adopter API unchanged.** Same call to fetch deferred SQL, same tracker to report status, same per-database semantics, same kill-switch / force-immediate / force-deferred config. Only visible difference: one fewer column in an internal table adopters don't query directly.

### Size

**−574 net lines across 35 files** vs. #375 (1 130 deletions, 556 insertions).
- Enricher: ~295 → ~170 lines
- `CreateDeployedIndexes` step: ~140 → ~85 lines
- `UpgradeContext` interface: **deleted**
- ~10 enricher unit tests removed (they tested the redundant checks)

### Phase breakdown

| SP | Scope |
|---|---|
| SP1 | Visitor gates tracking on `isDeferred()`; adds `service.prime()` for cross-session state |
| SP2 | Enricher rewritten: primes service + virtualizes unbuilt-deferred only; drops consistency checks |
| SP3 | Deletes `UpgradeContext`, 3-arg `execute`, prepopulation loop |
| SP4 | Documents why `collectDeferredIndexJobs` stays schema-scan (DAO approach has a timing bug) |
| SP5 | Drops `indexDeferred` column as redundant under slim invariant |
| SP6 | New slim-branch documentation |

## Test plan
- [ ] Review `TestDeployedIndexesIntegration` (25 tests) — deferred lifecycle on H2.
- [ ] Review slim enricher tests — priming + virtualization coverage.
- [ ] Compare `DeployedIndexesModelEnricherImpl` against the full-model version (#375) side-by-side; verify every deleted check is something `SchemaHomology` already catches.
- [ ] Confirm cross-upgrade carry-forward: upgrade A declares a deferred index; upgrade B renames its column; `getDeferredIndexStatements()` after B returns SQL with the renamed column.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


michal-morzywolek:
Closing — design has changed. The deferred-index work has converged on `experimental/deferred-indexes-background-build`; this branch is no longer being pursued.

PR #377, opened 2026-04-21
Use apache poi instead of jxl

jonknight73:
JXL last made a release in May 2011, it is abandonware. This PR replaces it with apache poi which is actively maintained.


jonknight73:
Replacing JXL with POI removed an accidental transitive source of Log4j 1.x used by Commons Logging during tests. The build already expects a Log4j-backed test logging setup, so Log4j is now declared explicitly as a shared test dependency rather than being inherited transitively.

PR #382, opened 2026-07-16
Make extension views in postgres be recognised as system views

jonknight73:
Make Extension views be recognised as system views so they are excluded from upgrade path calculations.

### DIFF
--- a/morf-postgresql/src/main/java/org/alfasoftware/morf/jdbc/postgresql/PostgreSQLMetaDataProvider.java
+++ b/morf-postgresql/src/main/java/org/alfasoftware/morf/jdbc/postgresql/PostgreSQLMetaDataProvider.java
@@ -5,6 +5,7 @@ import static org.alfasoftware.morf.jdbc.DatabaseMetaDataProviderUtils.getDataTy
 import static org.alfasoftware.morf.jdbc.DatabaseMetaDataProviderUtils.shouldIgnoreIndex;
 
 import java.sql.Connection;
+import java.sql.PreparedStatement;
 import java.sql.ResultSet;
 import java.sql.SQLException;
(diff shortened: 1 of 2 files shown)


morf-postgresql/src/main/java/org/alfasoftware/morf/jdbc/postgresql/PostgreSQLMetaDataProvider.java
```diff
@@ -57,6 +68,48 @@ protected boolean isPrimaryKeyIndex(RealName indexName) {
   }
 
 
+  @Override
+  protected boolean isSystemTable(RealName tableName) {
+    return extensionRelationNames.get().contains(tableName);
+  }
+
```
jsimlo, line 75: Calling super as well would be good pracitise :)

morf-postgresql/src/main/java/org/alfasoftware/morf/jdbc/postgresql/PostgreSQLMetaDataProvider.java
```diff
@@ -57,6 +68,48 @@ protected boolean isPrimaryKeyIndex(RealName indexName) {
...
+  @Override
+  protected boolean isSystemTable(RealName tableName) {
+    return extensionRelationNames.get().contains(tableName);
+  }
+
+
+  @Override
+  protected boolean isSystemView(RealName viewName) {
+    return extensionRelationNames.get().contains(viewName);
+  }
```
jsimlo, line 80: So there was something about "SystemView" before, that is good :)
Again, `or super`.

jsimlo (changes requested):

PR #384, opened 2026-08-25
Add hashing function

georgiasearle-dot:
Add new hashing function to facilitate obfuscation for a given database type.

### DIFF
--- a/morf-core/src/main/java/org/alfasoftware/morf/jdbc/SqlDialect.java
+++ b/morf-core/src/main/java/org/alfasoftware/morf/jdbc/SqlDialect.java
@@ -2785,7 +2785,7 @@ public abstract class SqlDialect {
 
   /**
    * Converts the HASH function into SQL.
-   * The hash produced will be all in lowercase.
+   * The hash produced will all be in lowercase.
    *
    * @param field the value to hash.
    * @param salt the salt to append before hashing. This field can be left as an empty string if no salt is required.
--- a/morf-core/src/main/java/org/alfasoftware/morf/jdbc/SqlDialect.java
+++ b/morf-core/src/main/java/org/alfasoftware/morf/jdbc/SqlDialect.java
@@ -2239,6 +2239,12 @@ public abstract class SqlDialect {
         }
         return getSqlForRowNumber();
 
+      case HASH:
+        if (function.getArguments().size() != 3) {
+          throw new IllegalArgumentException("The HASH function should have three arguments. This function has " + function.getArguments().size());
+        }
+        return getSqlForHash(function.getArguments().get(0), function.getArguments().get(1), function.getArguments().get(2));
+
       default:
         throw new UnsupportedOperationException("This database does not currently support the [" + function.getType() + "] function");
     }
@@ -2775,6 +2781,17 @@ public abstract class SqlDialect {
   protected String getSqlForRightPad(AliasedField field, AliasedField length, AliasedField character) {
     return "RPAD(" + getSqlFrom(field) + ", " + getSqlFrom(length) + ", " + getSqlFrom(character) + ")";
   }
+
+
+  /**
+   * Converts the HASH function into SQL.
+   *
+   * @param field the value to hash.
+   * @param salt the salt to append before hashing.
+   * @param length the length of the hash.
+   * @return string representation of the SQL.
+   */
+  protected abstract String getSqlForHash(AliasedField field, AliasedField salt, AliasedField length);
 
 
   /**
--- a/morf-core/src/main/java/org/alfasoftware/morf/jdbc/SqlDialect.java
+++ b/morf-core/src/main/java/org/alfasoftware/morf/jdbc/SqlDialect.java
@@ -2240,10 +2240,10 @@ public abstract class SqlDialect {
         return getSqlForRowNumber();
 
       case HASH:
-        if (function.getArguments().size() != 3) {
-          throw new IllegalArgumentException("The HASH function should have three arguments. This function has " + function.getArguments().size());
+        if (function.getArguments().size() != 2) {
+          throw new IllegalArgumentException("The HASH function should have two arguments. This function has " + function.getArguments().size());
         }
-        return getSqlForHash(function.getArguments().get(0), function.getArguments().get(1), function.getArguments().get(2));
+        return getSqlForHash(function.getArguments().get(0), function.getArguments().get(1));
 
       default:
         throw new UnsupportedOperationException("This database does not currently support the [" + function.getType() + "] function");
@@ -2785,13 +2785,13 @@ public abstract class SqlDialect {
 
   /**
    * Converts the HASH function into SQL.
+   * The hash produced will be all in lowercase.
    *
    * @param field the value to hash.
-   * @param salt the salt to append before hashing.
-   * @param length the length of the hash.
+   * @param salt the salt to append before hashing. This field can be left as an empty string if no salt is required.
    * @return string representation of the SQL.
    */
-  protected abstract String getSqlForHash(AliasedField field, AliasedField salt, AliasedField length);
+  protected abstract String getSqlForHash(AliasedField field, AliasedField salt);
 
 
   /**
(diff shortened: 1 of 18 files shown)


morf-core/src/main/java/org/alfasoftware/morf/sql/element/Function.java
```diff
@@ -658,13 +658,28 @@ public static Function rightPad(AliasedField field, int length, String character
...
+   * Helper method to create a hashing function.
+   *
+   * @param field the value to hash.
+   * @param salt the salt to append before hashing.
+   * @param length the length of the hash.
+   * @return an instance of the hash function.
+   */
+  public static Function hash(AliasedField field, AliasedField salt, AliasedField length) {
+    return new Function(FunctionType.HASH, field, salt, length);
+  }
```
Amarrd: If for whatever reason no salt is required, could this argument be provided as an empty string? Looks like the concat should handle this for the various dialects, but might be worth mentioning explicitly in the docs

morf-core/src/test/java/org/alfasoftware/morf/jdbc/MockDialect.java
```diff
@@ -308,6 +308,12 @@ protected String getSqlForNow(Function function) {
   }
 
 
+  @Override
+  protected String getSqlForHash(AliasedField field, AliasedField salt, AliasedField length) {
+    return "HASH(" + getSqlFrom(field) + "' " + getSqlFrom(salt) +")";
```
Amarrd: This "' " value doesn't look right. Can't see an equivalent TestMockDialect class to see how this is being used so maybe not important, but we should probably keep this consistent with one of the real dialects.

morf-oracle/src/main/java/org/alfasoftware/morf/jdbc/oracle/OracleDialect.java
```diff
@@ -1236,6 +1236,13 @@ protected String getSqlForNow(Function function) {
   }
 
 
+  @Override
+  protected String getSqlForHash(AliasedField field, AliasedField salt, AliasedField length) {
+    return String.format("SUBSTR(LOWER(RAWTOHEX(STANDARD_HASH(%s || %s, 'SHA256'))), 1, %s)",
```
Amarrd: Do the rest of the dialects return lowercase hashes so we've added the LOWER here to keep that consistent? If so it'll be worth making it clear in the docs that we'll always return a lowercase, or removing this and accepting the variation across databases

morf-core/src/main/java/org/alfasoftware/morf/sql/element/Function.java
```diff
@@ -658,13 +658,28 @@ public static Function rightPad(AliasedField field, int length, String character
   }
 
 
+  /**
+   * Helper method to create a hashing function.
+   *
+   * @param field the value to hash.
+   * @param salt the salt to append before hashing.
+   * @param length the length of the hash.
```
Amarrd: Should we include the length here? None of the underlying HASH functions for the dialects require it, and if the full output of the hash is required it's not clear how to achieve this. By doing a [SUBSTRING](https://github.com/alfasoftware/morf/blob/5f3c0085557e1fd6fa8b3a1e02be55a440d11828/morf-core/src/main/java/org/alfasoftware/morf/sql/element/Function.java#L291) on the hashed value we can get the same result for our obfuscation purposes and keep this function more tightly scoped - what do you think?

morf-core/src/main/java/org/alfasoftware/morf/sql/element/Function.java
```diff
@@ -658,13 +658,28 @@ public static Function rightPad(AliasedField field, int length, String character
   }
 
 
+  /**
+   * Helper method to create a hashing function.
+   *
+   * @param field the value to hash.
+   * @param salt the salt to append before hashing.
```
Amarrd: If for whatever reason no salt is required, could this argument be provided as an empty string? Looks like the concatenation should be handled for the various dialects, but worth confirming and mentioning explicitly in the docs

georgiasearle-dot: I imagine an empty String will work fine, but will definitely test it out and add to the docs

morf-core/src/main/java/org/alfasoftware/morf/sql/element/Function.java
georgiasearle-dot: Agreed it isn't necessary here. Will update

morf-oracle/src/main/java/org/alfasoftware/morf/jdbc/oracle/OracleDialect.java
georgiasearle-dot: Yes Oracle was returning a hash in all uppercase, and I wanted to keep the obfuscation consistent  (so  fields like billingAddressNameLongUpper will be the only uppercase values). I'll add this to the doc

Amarrd (dismissed):

morf-core/src/main/java/org/alfasoftware/morf/sql/element/Function.java
```diff
@@ -658,13 +658,28 @@ public static Function rightPad(AliasedField field, int length, String character
   }
 
 
+  /**
+   * Helper method to create a hashing function.
+   * The hash produced will all be in lowercase.
+   *
+   * @param field the value to hash.
+   * @param salt the salt to append before hashing. This field can be left as an empty string if no salt is required.
```
Amarrd:
```suggestion
   * @param salt the salt to append before hashing. This field can be provided as an empty FieldLiteral if no salt is required.
```

Amarrd (dismissed):

morf-core/src/main/java/org/alfasoftware/morf/jdbc/SqlDialect.java
```diff
@@ -2239,6 +2239,12 @@ protected String getSqlFrom(Function function) {
         }
         return getSqlForRowNumber();
 
+      case SHA256_HEX:
+        if (function.getArguments().size() != 2) {
+          throw new IllegalArgumentException("The HASH function should have two arguments. This function has " + function.getArguments().size());
```
jsimlo: `The HASH function`

morf-core/src/test/java/org/alfasoftware/morf/jdbc/MockDialect.java
```diff
@@ -308,6 +308,12 @@ protected String getSqlForNow(Function function) {
   }
 
 
+  @Override
+  protected String getSqlForSHA256Hex(AliasedField field, AliasedField salt) {
+    return "ENCODE(SHA256(CAST(" + getSqlFrom(field) + " || " + getSqlFrom(salt) + " as bytea)), 'hex')" ;
```
jsimlo: I would say something very simple here:
```
return "SHA256HEX(" + getSqlFrom(field) + " || " + getSqlFrom(salt) + ")";
```

It is a mock dialect, so it does not have to follow any real dialect, and this would be easy to understand.

morf-core/src/main/java/org/alfasoftware/morf/sql/element/Function.java
```diff
@@ -658,13 +658,28 @@ public static Function rightPad(AliasedField field, int length, String character
...
+   * Helper method to create a SHA-256 hashing function.
+   * The hash produced will all be an all lowercase hex string.
+   *
+   * @param field the value to hash.
+   * @param salt the salt to append before hashing. This field can be provided as an empty FieldLiteral if no salt is required.
+   * @return an instance of the hash function.
+   */
+  public static Function sha256Hex(AliasedField field, AliasedField salt) {
+    return new Function(FunctionType.SHA256_HEX, field, salt);
+  }
```
jsimlo, line 671: Since you have added that no-salt option, which requires a bit of internal knowledge, might be good to add `public static Function sha256Hex(AliasedField field) { return sha256Hex(field, literal('')); }` variant, which would help people to use sha256 hash without a salt

jsimlo (changes requested):

jsimlo (approved):

